PR #217, opened 2021-05-09
⚡ Add warm start capability to fva and pfba

stelmo:
Warm starts are a well documented way to increase the efficiency of solving consecutive optimization problems. I have added warm start capabilities to FVA and pFBA (it doesn't make sense for FBA, unless we make a new function that calls many of them on a cluster?).

The only issue is that I do not know which solvers support this, except for Gurobi... The open source ones tend not to I think...

### DIFF
--- a/src/analysis/flux_variability_analysis.jl
+++ b/src/analysis/flux_variability_analysis.jl
@@ -145,7 +145,8 @@ function _FVA_optimize_reaction(model, rid, ret)
     var = all_variables(model)[abs(rid)]
 
     @objective(model, sense, var)
-    if termination_status(model) != MOI.OPTIMIZE_NOT_CALLED && solver_name(model) in ["Gurobi", ]
+    if termination_status(model) != MOI.OPTIMIZE_NOT_CALLED &&
+       solver_name(model) in ["Gurobi"]
         set_start_value.(all_variables(model), value.(all_variables(model))) # warm start
     end
     optimize!(model)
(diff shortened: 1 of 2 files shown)


src/analysis/flux_variability_analysis.jl
```diff
@@ -145,6 +145,10 @@ function _FVA_optimize_reaction(model, rid, ret)
     var = all_variables(model)[abs(rid)]
 
     @objective(model, sense, var)
+    if termination_status(model) != MOI.OPTIMIZE_NOT_CALLED &&
+       solver_name(model) in ["Gurobi"]
```
exaexa, line 149: - is this really gurobi-only?
- wouldn't it be better for the warm start to come from the previous fba?

exaexa (changes requested):

src/analysis/flux_variability_analysis.jl
stelmo: I am not sure which solvers support this and which don't. Previously JuMP indicated that this happens automatically. I think doing it this was ensures that the previous solution is used? The solution is "contained" in `model` right?

exaexa: yeah, still technically there's probably no reason to make this work
- feasible solution is usually found very quickly and can be copied from the previous problem (if we want to optimize that, we should optimize JuMP/solvers, not COBREXA)
- caching the previous states won't help much because they are usually on a completely different side of the polytope

stelmo: yeah, I agree! closing this PR ⚰️

PR #251, opened 2021-05-11
Fix hit and run sampler

stelmo:
Fix hit and run sampler to be production ready


src/analysis/sampling/hit_and_run.jl
```diff
@@ -47,33 +47,70 @@ samples = hit_and_run(100_000, opt_model; keepevery=10, samplesize=5000)
...
+    ws, lbs, ubs = warmup(
+        model,
+        optimizer;
+        modifications = modifications,
+        warmup_points = warmup_indices,
+        workers = workers, # parallel
+    )
+
+    # load warmup points to workers
+    map(fetch, save_at.(workers, :ws, Ref(:($ws))))
```
exaexa, line 71: this can be done more easily in one step, you'd do something like:

```
save_at.(workers, :_cobrexa_hit_and_run_warmup, Ref(:($ws, $lbs, $ubs)))
```

btw please use the _cobrexa_prefixed_symbols to avoid name collisions (it's saved into the global namespace on remotes).

src/analysis/sampling/hit_and_run.jl
```diff
@@ -47,33 +47,70 @@ samples = hit_and_run(100_000, opt_model; keepevery=10, samplesize=5000)
...
+    )
 
-    wpoints = _get_warmup_points(opt_model; random_objective = random_objective)
+    # remove warmup points from workers
+    map(fetch, remove_from.(workers, :ws))
+    map(fetch, remove_from.(workers, :lbs))
+    map(fetch, remove_from.(workers, :ubs))
+
+    # not sure how to do this better
+    vals = zeros(samplesize, length(lbs), nchains)
```
exaexa, line 88: I guess something as in `vcat(samples...)` could help, e.g. like `cat(dim=3, samples...)`

stelmo: didn't work, no biggie though :)

stelmo:
@exaexa so here are the `almost` working sampler functions. Basically everything works in serial, but the wheels fall of as soon as I add any compute nodes on my pc. See the testing functions `test\analysis\sampling*.jl`. I think the problem originates in `warmup.jl`... When I remove

exaexa (approved):

PR #261, opened 2021-05-17
docs badges & acks

### LINKED ISSUE #250: add backlink from docs
**Describe the solution you'd like**
an easy clickthrough from docs main page to git repo

**Describe alternatives you've considered**
none, it would be best to find how/if others do it

**Additional context**
thanks @Trefex

exaexa:
cross-repo PR! :hammer: 

Fixes #250, also reacts to the PMC comments we got about all landing pages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,4 +131,4 @@ The development was supported by European Union's Horizon 2020 Programme under
 PerMedCoE project ([permedcoe.eu](https://www.permedcoe.eu/)) agreement no.
 951773.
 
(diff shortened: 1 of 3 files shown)


laurentheirendt (changes requested):
I think it would be best if we had the same ack's in `README.md` than in `index-template.md`

exaexa:
> I think it would be best if we had the same ack's in `README.md` than in `index-template.md`

I can cut it out as with the guide, the main problem is that the "main" part of the acks (the logos) are not "portable" b/c links will break (or should I just put absolute URLs there?)

laurentheirendt:
maybe then absolute URLs are best ...

laurentheirendt:
or we replace `assets/` properly during the documentation generation - I'd prefer that actually

laurentheirendt:
in consideration of #262, can you open from a branch from the upstream repo?

PR #263, opened 2021-05-17
add singularity container

laurentheirendt:
overwrites #260


docs/make.jl
```diff
@@ -70,5 +70,5 @@ deploydocs(
     repo = "github.com/$(ENV["TRAVIS_REPO_SLUG"]).git",
     target = "build",
     branch = "gh-pages",
-    push_preview = true
+    push_preview = true,
```
exaexa: I'm really really interested when this disappears :D :D

exaexa (approved):

docs/make.jl
laurentheirendt: it just did 😄

PR #265, opened 2021-05-17
switch README to Tulip

### LINKED ISSUE #232: change suggested solved in README from GLPK to Tulip
I think it is more appropriate to select Tulip as the suggested solver

exaexa:
Closes #232


stelmo (approved):
Go Julia :)

laurentheirendt (approved):

PR #269, opened 2021-05-18
Speed up convert to StandardModel by using threads

stelmo:
Converting models to `StandardModel` is pretty slow. I added a small improvement to use threads to process genes, reactions and metabolites on separate threads to speed this up. Slight improvements noted.

Results using iML1515 (2712 reactions, 1877 metabolites) as an example on my laptop using 4 threads (before, after timings):

iML1515.json: 7.2 seconds -> 5.4 seconds ~ 25% improvement
iML1515.mat: 1.4 seconds -> 1 second ~ 28% improvement
iML1515.xml: 5.6 seconds -> 5.5 seconds ~ 2% improvement

It seems to help for all the model types except SBML... No idea why it is different there...?

### DIFF
--- a/src/COBREXA.jl
+++ b/src/COBREXA.jl
@@ -12,6 +12,7 @@ using OrderedCollections
 using Random
 using SparseArrays
 using Statistics
+using Base.Threads
 
 import Base: findfirst, getindex, show
 import Pkg
--- a/src/base/types/StandardModel.jl
+++ b/src/base/types/StandardModel.jl
@@ -303,44 +303,53 @@ function Base.convert(::Type{StandardModel}, model::MetabolicModel)
     metids = metabolites(model)
     rxnids = reactions(model)
 
-    for gid in gids
-        modelgenes[gid] = Gene(
-            gid;
-            notes = gene_notes(model, gid),
-            annotations = gene_annotations(model, gid),
-        ) # TODO: add name accessor
-    end
-
-    for mid in metids
-        modelmetabolites[mid] = Metabolite(
-            mid;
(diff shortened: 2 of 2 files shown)


laurentheirendt (commented):
would this implementation also still work when I'd have only 1 thread?

exaexa:
For optimizations, we first need some detailed benchmark of where&how the performance is getting lost. I guess the most of the slowness comes from the (ugly) way of picking out the reaction IDs from json models, which can be fixed much more effectively than just with throwing additional cores at it. I'll get to it asap.

stelmo:
> would this implementation also still work when I'd have only 1 thread?

Yes, all the tests are relative to the one thread case, which is the same speed as the original code. However, let's talk about threads more with @exaexa before going further, I don't think he was super supportive of adding this kind of functionality it.

laurentheirendt (commented):
I am not sure that it is a good idea to have the `@timeit` macro calls inside the `convert` function by default

stelmo:
> I am not sure that it is a good idea to have the `@timeit` macro calls inside the `convert` function by default

Yep, I will remove these if we decide on going forward with this PR. @exaexa should we even bother with threads? If not, then we can delete this PR

exaexa:
Threads are better avoided here. Instead we should make the conversion efficient with the indexes. I'll make it work tomorrow.

stelmo:
Cool, closing this PR then

stelmo:
> Threads are better avoided here. Instead we should make the conversion efficient with the indexes. I'll make it work tomorrow.

Ah yeah, I promised to do that! Will start now!

PR #202, opened 2021-05-07
🚧 Make a community model tutorial

stelmo:
Let's see what happens?

PR #270, opened 2021-05-19
✨ Basic community model construction functions

stelmo:
Add some functions that will make community model construction easier.
1. `join` many single cell models together
2. `add_model` a single cell model to a community model
3. some helper functions

Basically during #202 I realized that creating community models is tricky and time consuming. Here I attempt to make functions that automate this process. They all use `CoreModel` since the models can get quite big. I tried to make the construction process as efficient as possible, especially with regard to the stoichiometric matrix construction part.

I tested `join` and it seems to scale linearly with model complexity (2 -> 2000 iML1515 models combined, max number of reactions 5424337):

```
                                       join scaling
                         ┌────────────────────────────────────────┐
                     300 │⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀│
                         │⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀│
                         │⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀│
                         │⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀│
                         │⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⣠│
                         │⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⣀⠤⠒⠉⠀│
                         │⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⣀⠤⠒⠉⠀⠀⠀⠀⠀│
   Time required [s]     │⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢀⣀⠤⠒⠉⠀⠀⠀⠀⠀⠀⠀⠀⠀│
                         │⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢀⡠⠔⠊⠁⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀│
                         │⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⡠⠔⠊⠁⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀│
                         │⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⣀⠤⠒⠉⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀│
                         │⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢀⠤⠒⠉⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀│
                         │⠀⠀⠀⠀⠀⠀⠀⢀⡠⠔⠊⠁⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀│
                         │⠀⠀⠀⢀⡠⠔⠊⠁⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀│
                       0 │⡠⠔⠊⠁⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀│
                         └────────────────────────────────────────┘
                         0                                     2000
                              Number of iML1515 models joined
```


stelmo:
> There's a potential issue with `all_boundaries`: it might pick up reactions that are not exchanges, for example "demand reactions".
> 
> See eg
> 
> ```julia
> agora=load_model(StandardModel, "test/data/agora-model.mat")
> excR, excM = all_boundaries(agora)
> excR[1]
> ```
> 
> It's not exactly ideal but check out `find_exchange_reactions` and `find_exchange_metabolites` in `src/reconstruction/CoreModel`.
> 
> edit. Another curiosity in this and some other models: it already has a biomass metabolite and an exchange reaction for that:
> 
> ```julia
> agora=load_model(StandardModel, "test/data/agora-model.mat")
> metabolites(agora)[142]
> stoichiometry(agora)[142, :]
> reactions(agora)[188]
> ```
> 
> This might also cause issues.

Thanks for looking at this code! So yes, both the things you pointed out are issues. For each point:

1. Yes, `all_boundaries` returns demands, sinks and exchanges. In the test code I just pushed I isolate those for further processing. The reason `all_boundaries` doesn't just return the exchanges by looking for "signatures" like "EX_" in the reaction name, is that the user could import a weird model where they name them differently. So identifying the exchange reactions is up to the user. Those functions you mention would do exactly that, although I currently just do:
```julia
boundary_rxn_ids, boundary_met_ids = all_boundaries(m2)
exchange_rxn_ids = filter(startswith("EX_"), boundary_rxn_ids)
exchange_met_ids = filter(endswith("_e"), boundary_met_ids)
```
2. Yes, agora does do exactly that. In that case the user would not use the objective keyword argument and do something like this:
```julia
models = [load_model(CoreModel, org) for org in agora_files]
for model in models
    remove_reactions(model, "EX_biomass(e)")
end

# find exchanges
...more-code...

# build community model
community = Evolution.joinmodels(models, exrxns, exmets; add_biomass_objective = false)

# find biomass indices
biomass_minds = findall(x->occursin("biomass", x), metabolites(community))

# add objective column
nr, _ = size(community.S)
objcol = spzeros(nr)
objcol[biomass_minds] .= -1.0

community.S = hcat(community.S, objcol)
community.xl = [community.xl; 0.0]
community.xu = [community.xu; 1000.0]
community.rxns = [community.rxns; "biomass"]
community.b = spzeros(size(community.S, 1))
community.c = spzeros(size(community.S, 2))
community.c[end] = 1.0
```
I am still working on getting this smoothed out and working nicely, currently the models joined together give rise to infeasible bounds. In the process of figuring it out, I will loudly announce on slack once this is sorted :)

However, I am still not sure about what to automate and what to leave to the user to do, e.g. in the agora case, should the join function be able to handle their format automatically or should the join literally just connect the stoichiometric matrices and leave the biomass functions/metabolites alone? Not sure...

src/io/show/MetabolicModel.jl
```diff
@@ -3,7 +3,7 @@ Pretty printing of everything metabolic-modelish.
 """
 function Base.show(io::IO, ::MIME"text/plain", m::MetabolicModel)
     _pretty_print_keyvals(io, "", "Metabolic model of type $(typeof(m))")
-    if prod(size(stoichiometry(m))) < 5_000_000
+    if prod(size(stoichiometry(m))) < 20_000_000
```
laurentheirendt, line 6: shall we not have this in the constants?

src/reconstruction/CoreModel.jl
```diff
@@ -200,6 +200,60 @@ function verify_consistency(
     return (new_reactions, new_metabolites)
 end
 
+"""
+    remove_metabolites(model::CoreModel, metabolites)
```
laurentheirendt, line 204: Will we also have `remove_metabolites!`? or do we have already?

src/reconstruction/community.jl
```diff
@@ -0,0 +1,388 @@
...
+            DomainError(
+                "exchange metabolite/reaction not found.",
+                "Exchange metabolites/reactions must already be contained in the community model.",
+            ),
+        )
+    end
+
+    n_cmodel_rows, n_cmodel_cols = size(stoichiometry(community))
+    n_model_rows, n_model_cols = size(stoichiometry(model))
+    Iadd, Jadd, Vadd = findnz(stoichiometry(model))
```
laurentheirendt, line 105: weird variable names

laurentheirendt: what is `I`, `J`, and `V`?

src/reconstruction/community.jl
```diff
@@ -0,0 +1,388 @@
...
+    n_reactions_total = n_cmodel_cols + n_model_cols
+
+    I, J, V = findnz(stoichiometry(community))
+    I = [I; Iadd]
+    J = [J; Jadd]
+    V = [V; Vadd]
+    S = sparse(I, J, V, n_metabolites_total, n_reactions_total)
+
+    lbsadd, ubsadd = bounds(model)
+    lbs, ubs = bounds(community)
```
laurentheirendt, line 139: wouldn't it be `lb` and `ub` only?

src/reconstruction/community.jl
```diff
@@ -0,0 +1,388 @@
...
+`emᵢ`, and these get linked to environmental exchanges, `erᵢ`. These `erᵢ`
+behave like normal single organism exchange reactions. When
+`add_biomass_objective` is true each model's biomass becomes a pseudo-metabolite
+(`bmᵢ`). These can be weighted in column `b`, called the `community_biomass`
+reaction in the community model, if desired. Refer to the tutorial if this is
+unclear.
+
+# Example
+```
+model_path = joinpath("..","models","e_coli_core.json")
```
laurentheirendt, line 216: I still believe that we should have doctests ...

laurentheirendt (changes requested):
nice work 👍🏼

src/reconstruction/community.jl
stelmo: yeah, these are the default variable names used by SparseArrays from Julia, I figured we should stick to their defaults to make it simpler for someone else to read? But basically, a sparse matrix can be represented by a set of 3 arrays. Suppose `M` is some sparse matrix, then `I` = row indices, `J` = column indices and `V` = values at the associated indices. So `I[a] = i`, `J[a]=j` and then `M[i,j] = V[a]`

src/reconstruction/community.jl
stelmo: So these are arrays of bounds e.g. a bound for each reaction i.e. `lbs = [lb_1, lb_2, ...]` hence the plural

src/reconstruction/community.jl
stelmo: easily done! How do I load "e_coli_core.json" for the doctests?

src/io/show/MetabolicModel.jl
stelmo: done!

src/reconstruction/CoreModel.jl
laurentheirendt: yes, please make an issue

src/reconstruction/community.jl
laurentheirendt: can you add this to the comments?

src/reconstruction/community.jl
laurentheirendt: can you add a comment?

src/reconstruction/community.jl
laurentheirendt: in the doctests.jl file

src/reconstruction/community.jl
stelmo: done

src/reconstruction/community.jl
stelmo: done

src/reconstruction/community.jl
stelmo: done

src/reconstruction/CoreModel.jl
```diff
@@ -200,6 +200,60 @@ function verify_consistency(
...
+```
+model = load_model(CoreModel, "e_coli_core.json")
+
+m1 = remove_metabolites(model, ["glc__D_e", "for_c"])
+m2 = remove_metabolites(model, "glc__D_e")
+m3 = remove_metabolites(model, indexin(["glc__D_e", "for_c"], metabolites(model)))
+m4 = remove_metabolites(model, first(indexin(["glc__D_e"], metabolites(model))))
+```
+"""
+function remove_metabolites(model::CoreModel, mets)
```
exaexa, line 222: this is technically a variant (except for argument order), we'd also need to have a ! variant which would be more appropriate in this case.

test/doctests.jl
```diff
@@ -1,9 +1,15 @@
 using Test, Documenter, COBREXA
 
 ex = quote
-    using COBREXA, Tulip
+    using COBREXA, Tulip, Downloads
     include(joinpath(dirname(pathof(COBREXA)), "..", "test", "data", "test_models.jl"))
     model = test_LP()
+    core_model_path =
+        joinpath(dirname(pathof(COBREXA)), "..", "test", "data", "e_coli_core.json")
+    Downloads.download(
```
exaexa, line 9: use the download function from tests, ideally declare this variable in test_models.jl (or have doctest_models.jl) so that this code stays clean.

src/reconstruction/community.jl
```diff
@@ -0,0 +1,409 @@
...
+
+"""
+    all_boundaries(model::M) where {M<:MetabolicModel}
+
+Return `boundary_rxns` and `boundary_mets` given `model`. Boundary reactions
+are reactions with a single stoichiometric coefficient in a column in the
+stoichiometric matrix, and boundary metabolites are the corresponding row/metabolite
+for that column.
+"""
+function all_boundaries(model::M) where {M<:MetabolicModel}
```
exaexa, line 394: name: `boundary_rxns_mets` or something. Declare the return type. You don't need the explicit qualification with M, just use ::MetabolicModel.

src/reconstruction/community.jl
```diff
@@ -0,0 +1,409 @@
...
+)
+
+size(stoichiometry(community))
+
+# output
+
+(166, 211)
+```
+"""
+function Base.join(
```
exaexa, line 246: the name `join` may be a bit too generic here, moreover it's in a base. It doesn't say that we're joining the models in a way that makes a community with exchange reactions, and there might be other ways.

Better ideas:
- `make_community_with_exchanges`
- `join_with_exchanges`

src/reconstruction/community.jl
```diff
@@ -0,0 +1,409 @@
...
+        community.xl[objective_column_index] = 0.0
+        community.xu[objective_column_index] = 1000.0
+    end
+end
+
+add_objective!(model::CoreModel, objective_met::String, objective_weight::Float64) =
+    add_objective!(model, [objective_met], [objective_weight])
+
+"""
+add_model(
```
exaexa, line 60: spawning `M` into the type is not required, you don't need to unify it anywhere. Just do ::MetabolicModel.

Do not use empty strings (empty string is a perfectly valid ID) -- you want `nothing`.

Include the return type.

The thing soudns a bit like it could make a modification but the semantics (and arguments) seems more like inplace, choose either of that.

exaexa: (I'd vote for inplace here)

src/reconstruction/community.jl
```diff
@@ -0,0 +1,409 @@
...
+for that column.
+"""
+function all_boundaries(model::M) where {M<:MetabolicModel}
+    boundary_mets = String[]
+    boundary_rxns = String[]
+    S = stoichiometry(model)
+    rxns = reactions(model)
+    mets = metabolites(model)
+    for i = 1:size(S, 2)
+        j, b = findnz(S[:, i])
```
exaexa, line 401: use `nnz` or something similar to avoid spawning the long vectors

src/reconstruction/CoreModel.jl
exaexa: (for variant, you'd need to uncurry the arguments so that it can be used as `remove_metabolites("met123")(model)` )

exaexa: also, properly naming `remove_metabolites` vs `remove_metabolite` can help a lot with untangling ambiguous cases

exaexa (changes requested):

src/io/show/MetabolicModel.jl
```diff
@@ -3,7 +3,7 @@ Pretty printing of everything metabolic-modelish.
 """
 function Base.show(io::IO, ::MIME"text/plain", m::MetabolicModel)
     _pretty_print_keyvals(io, "", "Metabolic model of type $(typeof(m))")
-    if prod(size(stoichiometry(m))) < 5_000_000
+    if prod(size(stoichiometry(m))) < _constants.default_stoichiometrix_view_size
```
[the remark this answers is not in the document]
stelmo: actually, it wasn't problematic, the previous bound was too small i.e. when I joined two e coli models it didn't display... The graphical output is kinda nice though, I would not delete it

src/reconstruction/community.jl
stelmo: done!

src/reconstruction/community.jl
stelmo: done!

src/reconstruction/CoreModel.jl
stelmo: This will be addressed in a feature PR though, see issue #281

test/doctests.jl
stelmo: So I removed the doctest for the model joining function because it is difficult to test properly with the pretty printing, I have however added the files as you suggested because in a future PR we will add doctests for other functions that rely on the models being available.

src/reconstruction/community.jl
stelmo: Hmm this function is not really in place... It deconstructs everything and then makes a new model. I am not sure how efficient and in place version will be given that you need to extend all the matrices (you may as well just make a new one)?

Or it can be made trivially in place by just overwriting the community input model?

src/io/show/MetabolicModel.jl
exaexa, line 6: coult we have something more like `maximum(size(stoichiometry(m))) < 500`
Also, this kinda creates the matrix and then throws it away, might be better to call n_reactions() and n_metabolites()

exaexa: (also can we please have this in a separate PR?)

test/docttest_models.jl
```diff
@@ -0,0 +1,5 @@
+core_path = download_data_file(
```
exaexa, line 1: filename spelling

test/reconstruction/community.jl
```diff
@@ -0,0 +1,160 @@
+@testset "Small model join" begin
+    model_path = download_data_file(
```
exaexa, line 2: there's a separate file for unifying all downloads now

test/reconstruction/community.jl
```diff
@@ -0,0 +1,160 @@
...
+        ),
+    )
+
+    d = flux_balance_analysis_dict(community, Tulip.Optimizer)
+    @test size(stoichiometry(community)) == (166, 211)
+    @test isapprox(d["community_biomass"], 0.41559777495618294, atol = TEST_TOLERANCE)
+end
+
+@testset "Heterogenous model join" begin
+    iJO1366_mat = download_data_file(
```
exaexa, line 46: separate file for downloads

test/doctests.jl
```diff
@@ -1,9 +1,11 @@
 using Test, Documenter, COBREXA
 
 ex = quote
-    using COBREXA, Tulip
+    using COBREXA, Tulip, Downloads
     include(joinpath(dirname(pathof(COBREXA)), "..", "test", "data", "test_models.jl"))
     model = test_LP()
```
exaexa, line 6: I guess doctests should be able to live for quite a long time with the same files as found now in `data_downloaded.jl`, right?

exaexa (changes requested):
Please separate into 3 PRs:
- doctest preparation (does the new code run doctests?)
- prettyprinting fixes (I guess there may be some extra discussion)
- the actual joining of models

stelmo:
> Please separate into 3 PRs:
> 
> * doctest preparation (does the new code run doctests?)
> * prettyprinting fixes (I guess there may be some extra discussion)
> * the actual joining of models

1) I removed the doctests for these functions. No other function has doctests yet, let's do that in another PR.
2) I reverted the pretty printing fixes and will make a new PR for that... hmm maybe #300?
3) This is mostly model joining now, although there are some modifications I introduced into `CoreModel` as well

src/reconstruction/community.jl
```diff
@@ -0,0 +1,406 @@
...
+column corresponding to the `biomass_id` reaction then produces this new biomass
+metabolite with unit coefficient. Note, `exchange_rxn_ids` and
+`exchange_met_ids` must already exist in the `community` model.
+
+# Example
+```
+community = add_model(community, model, exchange_rxn_ids, exchange_met_ids; model_name="species_2", biomass_id="BIOMASS_Ecoli_core_w_GAM")
+```
+"""
+function add_model(
```
exaexa, line 82: cosmetic: rename to `add_model_with_exchanges` to match the other function

src/reconstruction/community.jl
```diff
@@ -0,0 +1,406 @@
...
+    # = row indices, J = column indices and V = values at the associated
+    # indices. So I[a] = i, J[a]=j and then M[i,j] = V[a]
+    Iadd, Jadd, Vadd = findnz(stoichiometry(model))
+
+    # shift to fit into community         
+    Iadd .+= n_cmodel_rows
+    Jadd .+= n_cmodel_cols
+
+    # when adding a single model not that many reactions, push! okay?
+    exchange_rxn_model_inds = indexin(exchange_rxn_ids, reactions(model))
```
exaexa, line 115: cosmetic naming: use `_idxs` instead of `_inds`

src/reconstruction/community.jl
```diff
@@ -0,0 +1,406 @@
...
+
+    # # build diagonal environmental exchanges
+    for i = 1:length(exchange_rxn_ids)
+        I[kstart] = n_metabolites_metabolic + i
+        J[kstart] = n_reactions_metabolic + i
+        V[kstart] = -1.0
+        kstart += 1
+    end
+
+    if add_biomass_objective
```
exaexa, line 334: shouldn't it be called `set_biomass_objective`? it's technically not _adding_ anything.

src/reconstruction/community.jl
```diff
@@ -0,0 +1,406 @@
...
+
+    S = sparse(
+        I[1:kstart-1],
+        J[1:kstart-1],
+        V[1:kstart-1],
+        n_metabolites_total,
+        n_reactions_total,
+    ) # could be that some microbes don't have all the exchanges, hence kstart-1
+
+    reaction_cumsum = cumsum(reaction_lengths)
```
exaexa, line 353: better name: `_starts` or `_offsets` (preferable)

src/reconstruction/community.jl
```diff
@@ -0,0 +1,406 @@
...
+    mets = metabolites(model)
+    for i = 1:size(S, 2)
+        n = nnz(S[:, i])
+        if n == 1
+            j, _ = findnz(S[:, i])
+            push!(boundary_mets, mets[first(j)])
+            push!(boundary_rxns, rxns[i])
+        end
+    end
+    return boundary_rxns, boundary_mets
```
exaexa, line 405: perhaps better return a named tuple here so that people don't need to bind this explicitly or pick it up using [1] [2]

src/reconstruction/community.jl
```diff
@@ -0,0 +1,406 @@
...
+reaction in the community model, if desired. Refer to the tutorial if this is
+unclear.
+
+# Example
+```
+m1 = load_model(core_model_path)
+m2 = load_model(CoreModel, core_model_path)
+
+boundary_rxn_ids, boundary_met_ids = boundary_reactions_metabolites(m2)
+exchange_rxn_ids = filter(startswith("EX_"), boundary_rxn_ids)
```
exaexa, line 227: you might want to demo the looks-like-exchange predicate here

exaexa (changes requested):
looks good, some tiny stuff to fix, mostly naming

src/reconstruction/community.jl
stelmo: yep, updating!

src/reconstruction/community.jl
stelmo: actually it does get added to matrix here

src/reconstruction/community.jl
stelmo: or maybe I should remove this function? with `looks_like` this isn't necessary anymore... what do you think?

htpusa: We could have another `looks_like` for boundaries based on stoichiometry?

stelmo: I don't think it should be called `looks_like` in that case since you can directly test if it is a boundary reaction (they only have one metabolite?).  That might be a good function to include somewhere in `src/base/util` though. For now I have removed this function

exaexa (approved):
ok cool, let's see this. :D

PR #273, opened 2021-05-23
fix a problem in screening

exaexa:
Reported by @stelmo 

How come this didn't break with single-worker tests?


stelmo (approved):

stelmo:
> Reported by @stelmo
> 
> How come this didn't break with single-worker tests?

It works!! 💯

PR #276, opened 2021-05-25
disable the edit link in docs

### LINKED ISSUE #245: fix `Edit on Github` button on index.md
The link points to index.md although it doesn't exist at it is auto-generated

exaexa:
It's a bit confusing anyway for this use-case (if you're reading the
documentation you usually don't want to edit it).

Side effect:
Closes #245


laurentheirendt (changes requested):
this will remove the link completely - I'd think that having the link there is actually quite handy ...

exaexa:
There's unfortunately nothing to configure it that way, we'd need to go with custom Documenter.

laurentheirendt:
Hm, there must be a way 😄  I'll have a look in the coming days

exaexa:
ok well, linking the relevant parts of code:
https://github.com/JuliaDocs/Documenter.jl/blob/master/src/Writers/HTMLWriter.jl#L1081

this might be a good target for monkeypatching, but likely incomplete:
https://github.com/JuliaDocs/Documenter.jl/blob/c942201b745abbdf155c880e1231cb8240507b41/src/Utilities/Utilities.jl#L491

laurentheirendt (approved):

PR #284, opened 2021-05-27
Clean up hit-and-run sampling

exaexa:
This includes a rebase to master, combining it with the original PR (#251) seems to be a merging overkill.


exaexa:
PS. yhank you again, good cylon-x.

src/analysis/sampling/affine_hit_and_run.jl
```diff
@@ -0,0 +1,138 @@
...
+
+There are total `chains` of hit-and-run runs, each on a batch of
+`size(warmup_points, 2)` points. The runs are scheduled on `workers`, for good
+load balancing `chains` should be ideally much greater than `length(workers)`.
+
+Each run continues for `maximum(sample_iters)` iterations; the numbers in
+`sample_iters` represent the iterations at which the whole "current" batch of
+points is collected for output. For example, `sample_iters=[1,4,5]` causes the
+process run for 5 iterations, returning the sample batch that was produced by
+1st, 4th and last (5th) iteration.
```
stelmo, line 25: I don't understand this? Is this keeping every first, 4th and 5th sample?

stelmo (approved):
Except for being slightly confused about the arguments to affine_hit_and_run, this looks super good :)

src/analysis/sampling/affine_hit_and_run.jl
exaexa: yeah, this is kinda explicit way how to nicely express the stuff that would be otherwise specified in something like `collect_every_nth_iter` with a lot of extra logic.

Here you can do:
- `sample_iters = 100:105`  = do 100 iters, then collect 5 ones
- `sample_iters = 100.+(1:5)` (the same)
- `sample_iters = 10 .* (1:3)` = collect iterations 10,20,30
etc

exaexa: btw I think the doc is clear but I may not see the problem, any rewording suggestions welcome

PR #289, opened 2021-05-28
Clean up test data handling

exaexa:
Major changes:

- there's now separate `tmpfiles` and `downloads` (maybe that should all go to `tempdir()` ?)
- gitignore seems to be a more happy place
- some files got renamed
- all binary blobs (except for perhaps `favicon.ico`) should be gone now

PR #290, opened 2021-05-28
Missing functions for `CoreModelCoupled`

htpusa:
# Most `CoreModel` functions are missing for `CoreModelCoupled`

I added the missing functions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 *.json
 *.mat
 *.xml
+*.smod
 
 .DS_Store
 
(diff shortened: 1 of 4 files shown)


exaexa (changes requested):
Code looks good to me, except it would be great to have docstrings that include the full function header (otherwise there's no telling from documentation about the parameter types of stuff).

Minor thing: the "ignore smod" is basically obsolete now because all testing data got a bit reorganized (see #289). But don't worry about that, I will handle that on merging.

laurentheirendt:
closing in favor of #307

PR #292, opened 2021-06-01
Small fixes

exaexa:
- make tests work again
- @stelmo can you check whether the gitignore works for you like this?


stelmo (approved):
yeah, the gitignore issues were more of a stelmo specific issue :)

PR #295, opened 2021-06-02
change singularity base img

laurentheirendt:


laurentheirendt:
For now, the Singularity container `.sif` files are hosted here: https://cloud.sylabs.io/library/cylon-x

PR #297, opened 2021-06-05
✨ Add `looks_like_xxx` functions to make finding specific reaction/metabolite types easy

stelmo:
~~`CoreModel` has these functions, and they are useful so I added them to `StandardModel`. Also moved some code around to make the location more obvious.~~

Make functions that can be used in `findall` and `filter` to find specific reaction/metabolite types. Also add accessors to get the reaction equation from a model.


src/base/utils/CoreModel.jl
```diff
@@ -17,3 +17,46 @@ Base.isequal(model1::CoreModelCoupled, model2::CoreModelCoupled) =
...
+
+"""
+Returns indices of exchange reactions.
+Exchange reactions are identified based on most commonly used prefixes.
+"""
+function find_exchange_reactions(
+    model::CoreModel;
+    exclude_biomass = false,
+    biomass_str::String = "biomass",
+    exc_prefs = ["EX_"; "Exch_"; "Ex_"; "R_EX_"],
```
htpusa, line 29: Should we include all versions with the added "R_"? Also, maybe this list could just be put in `constants.jl`?

stelmo: ah, yeah we probably should. I will do it now. @exaexa is going to hate this elaborate guessing game :)

htpusa: As long as he's hating the game and not the players...

src/base/utils/CoreModel.jl
```diff
@@ -17,3 +17,46 @@ Base.isequal(model1::CoreModelCoupled, model2::CoreModelCoupled) =
...
+)
+    is_exc = falses(n_reactions(model))
+    for pref in exc_prefs
```
stelmo, line 32: I have no idea why this works...

src/base/utils/CoreModel.jl
```diff
@@ -17,3 +17,46 @@ Base.isequal(model1::CoreModelCoupled, model2::CoreModelCoupled) =
...
+        is_exc = is_exc .| startswith.(model.rxns, pref)
+    end
+    exc_inds = findall(is_exc)
+    if exclude_biomass
```
stelmo, line 36: :bug: original code did not actually find the biomass reaction

stelmo:
Note, I changed the `CoreModel` find functions a little bit. They resemble the `StandardModel` functions more now. @exaexa lmk what you think?

src/base/utils/CoreModel.jl
```diff
@@ -17,3 +17,77 @@ Base.isequal(model1::CoreModelCoupled, model2::CoreModelCoupled) =
...
+    find_exchange_reactions(
+        model::CoreModel;
+        exclude_biomass = false,
+        biomass_strings = _constants.biomass_strings,
+        ex_prefixes = _constants.exchange_prefixes,
+    )::Vector{String}
+
+Returns indices of exchange reactions. Exchange reactions are identified based
+on most commonly used prefixes, these prefixes can be set using `ex_prefixes`.
+See `_constants.biomass_strings` for the default list. The biomass reaction is
```
htpusa, line 31: `_constants.exchange_prefixes` right?

src/base/utils/CoreModel.jl
```diff
@@ -17,3 +17,77 @@ Base.isequal(model1::CoreModelCoupled, model2::CoreModelCoupled) =
...
+also added if it can be identified by looking for the strings listed in
+`biomass_strings` in the reaction ids. If `exclude_biomass` is true then this
+does not occur.
+
+Note: biomass exchange reactions are counted as exchange reactions and will NOT
```
htpusa, line 36: So my original intention was to exclude a biomass exchange (and not pickup a biomass reaction in any case). Any reason to change this?

src/base/utils/CoreModel.jl
stelmo: yes, good spot!

src/base/utils/CoreModel.jl
stelmo: hmm not really, fixing now

stelmo:
I made these functions generic. This will be useful in the future since they will likely be used a lot with the community functions

src/base/utils/looks_like.jl
```diff
@@ -0,0 +1,34 @@
...
+    biomass_strings = _constants.biomass_strings,
+    ex_prefixes = _constants.exchange_prefixes,
+)   
+    if any(startswith(rxn_id, x) for x in ex_prefixes) && !any(occursin(x, rxn_id) for x in biomass_strings) # found exchange reaction
+        return true
+    elseif !exclude_biomass && any([occursin(x, rxn_id) for x in biomass_strings]) # biomass
+        return true
+    end
+    return false
+end
```
htpusa, line 34: I still don't get why we would want to use this to find biomass reactions? Now, if a user would input a model that has a biomass exchange ("EX_biomass") and calls with `exclude_biomass = false`, they would get both the biomass exchange and the biomass reaction, right?

stelmo: Yeah, that is how it currently works. Ah, I am confused about the behaviour you want? Do you want `exclude_biomass` to exclude biomass reactions with `"EX_"` only? and not give the normal biomass function?

htpusa: Yeah, that's exactly my original intent. Basically, I wanted to find exchange reactions but on the off chance that there is a biomass exchange in the model, to be able to exclude that. If it's confusing, we can leave biomass out of it completely and I'll just do the filtering afterwards.

stelmo: Now we also have a `looks_like_biomass_reaction` function :)

src/base/utils/looks_like.jl
```diff
@@ -0,0 +1,88 @@
...
+```
+filter(looks_like_exchange_reaction, reactions(model)) # returns strings
+findall(looks_like_exchange_reaction, reactions(model)) # returns indices
+
+# to use the optional arguments you need to instantiate the function
+filter(x -> looks_like_exchange_reaction(x; exclude_biomass=true), reactions(model)) # returns strings
+findall(x -> looks_like_exchange_reaction(x; exclude_biomass=true), reactions(model)) # returns indices
+```
+"""
+function looks_like_exchange_reaction(rxn_id::String;
```
exaexa, line 26: simplify logic, now the whole condition looks like, schematically, after removing the redundant returns:
```
if eb
	anyprefix && !anybiomass
else
	anyprefix
```

you may convert the if to logic:
```
(eb && anyprefix && !anybiomass)
 ||
(!eb && anyprefix)
```
and factor out the anyprefix:
```
anyprefix && (eb && !anyboimass  || !eb)
```

further, the `eb &&` in the parens is now redundant because either the whole paren is true (`!eb`) or doesn't depend on value of `eb` (`!!eb`). So finally,

```
anyprefix && (!eb || !anybiomass)
```

but whoa, this may shortcut better:

```
anyprefix && !(eb && anybiomass)
```

src/base/utils/looks_like.jl
```diff
@@ -0,0 +1,88 @@
...
+# Example
+```
+filter(looks_like_biomass_reaction, reactions(model)) # returns strings
+findall(looks_like_biomass_reaction, reactions(model)) # returns indices
+```
+"""
+function looks_like_biomass_reaction(rxn_id::String;
+    biomass_strings = _constants.exchange_suffixes,
+)::Bool
+    any(occursin(x, rxn_id) for x in biomass_strings) && return true
```
exaexa, line 61: simplify -- `if (x) return true else return false` == `return x`

src/base/utils/looks_like.jl
```diff
@@ -0,0 +1,88 @@
...
+# Example
+```
+filter(looks_like_exchange_metabolite, metabolites(model)) # returns strings
+findall(looks_like_exchange_metabolite, metabolites(model)) # returns indices
+```
+"""
+function looks_like_exchange_metabolite(met_id::String;
+    ex_suffixes = _constants.exchange_suffixes,
+)::Bool
+    any(endswith(met_id, x) for x in ex_suffixes) && return true
```
exaexa, line 86: same here

src/base/types/CoreModel.jl
```diff
@@ -93,6 +93,26 @@ function objective(a::CoreModel)::SparseVec
     a.c
 end
 
+"""
+    reaction_equation(model::CoreModel, rxn_id::String)::Dict{String, Float64}
+
+Return the reaction equation of reaction with id `rxn_id` in model. The reaction
+equation maps metabolite ids to their stoichiometric coefficients.
+"""
+function reaction_equation(m::CoreModel, rxn_id::String)::Dict{String, Float64}
```
exaexa, line 102: this is precisely called "reaction stoichiometry" in SBML,  I'd suggest we use the same naming.

Also, can we add this to the abstract accesors?

test/reconstruction/CoreModel.jl
```diff
@@ -1,51 +1,3 @@
-@testset "Find exchange reactions and metabolites" begin
```
exaexa, line 1: this got completely replaced?

exaexa (changes requested):

src/base/utils/looks_like.jl
stelmo: I want to heart this again ❤️

test/reconstruction/CoreModel.jl
stelmo: ah no, it got moved to `test/utils/looks_like.jl`

src/base/types/abstract/MetabolicModel.jl
```diff
@@ -306,6 +306,26 @@ function reaction_subsystem(model::MetabolicModel, reaction_id::String)::Maybe{S
...
 end
 
+"""
+    reaction_stoichiometry(model::MetaboliteModel, rxn_id::String)::Dict{String, Float64}
+
+Return the reaction equation of reaction with id `rxn_id` in model. The reaction
+equation maps metabolite ids to their stoichiometric coefficients.
+"""
+function reaction_stoichiometry(m::MetabolicModel, rxn_id::String)::Maybe{Dict{String, Float64}}
+    return nothing
```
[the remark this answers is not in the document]
stelmo: okay, I added a default implementation instead

test/base/types/abstract/MetabolicModel.jl
```diff
@@ -6,7 +6,7 @@ end
 @testset "Base abstract model methods require proper minimal implementation" begin
     @test_throws MethodError reactions(123)
     x = FakeModel(123)
-    for m in [reactions, metabolites, stoichiometry, bounds, objective]
+    for m in [reactions, metabolites, stoichiometry, bounds, objective, reaction_stoichiometry]
```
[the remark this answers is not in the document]
stelmo: removed this

stelmo:
done! @exaexa

Issue #281, opened 2021-05-27
Homogenize add/remove functions and add in place versions


stelmo:
Currently `CoreModel` has `remove_reactions`, `add_reactions` etc. and `StandardModel` has `add!`, and `rm!`. Convert the `StandardModel` functions into their `CoreModel` counterparts to keep the names of the functions the same. 

Also, add in place versions of these functions where possible.

htpusa:
There is also the `verify_consistency` functionality in `CoreModel` which is currently unfinished. The goal was to check, when adding a reaction, if the reaction is already in the model, and if so, if the definitions agree.

Personally I'd like to have this, mainly for checking if two reactions with the same id are truly the same. However, it needs to be decided what the desired behaviour would be (eg, different id, same stoichiometry, what do).

stelmo:
From discussion with @exaexa:

make these functions generic e.g.
```julia
remove_metabolites(mets) -> (model::CoreModel) -> remove_metabolites(model, mets)
remove_metabolites(mets) -> (model::StandardModel) -> remove_metabolites(model, mets)
 ```
method ambiguity. Make generic instead and then specialize on model type.

stelmo:
From #197: 
Currently removing reactions or metabolites breaks the model. Making removals work.
Be smart: rm_metabolites!(model,metabolites(model)[rowCount(stoichiometry(model) .!= 0).>0])

stelmo:
> From #197:
> Currently removing reactions or metabolites breaks the model. Making removals work.
> Be smart: rm_metabolites!(model,metabolites(model)[rowCount(stoichiometry(model) .!= 0).>0])

actually, an informative error function in `stoichiometry` is supplied - this should alert the user. probably not required to fix now.

PR #298, opened 2021-06-05
🚚 Rename `add!` and `rm!` functions for `StandardModel`

stelmo:
Currently `CoreModel` uses e.g. `remove_reaction` and `add_reaction` etc. However, `StandardModel` uses `add!` and `rm!`, this adds too much variability - the same method names should apply. Here `StandardModel`'s functions are renamed. 

However, a slight issue remains that the functions are in place for `StandardModel`, hence they have the bang!, while for `CoreModel` they make a new model and are thus not in place. The user thus needs to remember to use the bang for `StandardModel`s and not for `CoreModel`s. This is slightly non-optimal... what do you think?

### DIFF
--- a/src/reconstruction/StandardModel.jl
+++ b/src/reconstruction/StandardModel.jl
@@ -5,18 +5,11 @@ Add `rxns` to `model` based on reaction `id`.
 """
 function add_reactions!(model::StandardModel, rxns::Vector{Reaction})
     for rxn in rxns
-        add_reaction!(model, rxn)
+        model.reactions[rxn.id] = rxn
     end
 end
 
-"""
-    add_reaction(model::StandardModel, rxn::Reaction)
-
-Add `rxn` to `model` based on reaction `id`.
-"""
-function add_reaction!(model::StandardModel, rxn::Reaction)
-    model.reactions[rxn.id] = rxn
-end
+add_reactions!(model::StandardModel, rxn::Reaction) = add_reactions!(model, [rxn])
 
 """
     add_metabolites!(model::StandardModel, mets::Vector{Metabolite})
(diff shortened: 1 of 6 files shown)


stelmo:
closes #281

src/reconstruction/StandardModel.jl
```diff
@@ -1,47 +1,41 @@
...
-        add!(model, met)
+        model.metabolites[met.id] = met
     end
 end
 
-function add!(model::StandardModel, met::Metabolite)
-    model.metabolites[met.id] = met
-end
-
+add_metabolites!(model::StandardModel, met::Metabolite) =  add_metabolites!(model, [met])
```
exaexa, line 25: ey, can we reduce the polymorphism here and call it just `add_metabolite!` ?

test/analysis/knockouts.jl
```diff
@@ -1,20 +1,20 @@
 @testset "single_knockout" begin
     m = StandardModel()
-    add!(m, Metabolite("A"))
-    add!(m, Metabolite("B"))
+    add_metabolites!(m, Metabolite("A"))
```
exaexa, line 3: (here the singular `add_metabolite` would look much more reasonable)

(nitpicking)

exaexa (approved):

src/reconstruction/StandardModel.jl
stelmo: lol, I actually had this but then second guessed myself. Will do!

test/analysis/knockouts.jl
stelmo: yep, I also think so. The only reason I had it like this is because in `CoreModel` this happens: 
```julia
function remove_reactions(m::CoreModel, rxn::Integer)
    return remove_reactions(m, [rxn])
end
```

hmm, we should pick one style though... I don't mind too much, you decide?

stelmo: went with singular/plural everywhere

Issue #288, opened 2021-05-27
Improve coverage of src/base/utils/StandardModel.jl


exaexa:
https://codecov.io/gh/LCSB-BioCore/COBREXA.jl/src/master/src/base/utils/StandardModel.jl

The `verbose` branch in `exchange_reactions` is completely untested, I'd either :axe: it (if the user needs a more complete output, then the actual output should be more complete, not "printed side effects". If the verbose variant is to be kept, consider using the standard per-category "debug" logging functions.

stelmo:
See PR #299

PR #299, opened 2021-06-05
🪓 Remove `exchange_reactions`

stelmo:
This function is unnecessary after #297 is merged. The user can just find the exchanges themselves and display/use the flux dict themselves... The most useful part of this was to "pretty print" the result of some analysis function... which might be a good idea... but this will be for another PR.


stelmo:
This closes #288

exaexa (approved):

PR #300, opened 2021-06-05
💄 Print flux solutions in a nicer/more useful way

stelmo:
When the flux solution is returned as a dictionary the display can be improved from:
```julia
julia> sol = flux_balance_analysis_dict(model, Tulip.Optimizer)
Dict{String, Float64} with 95 entries:
  "ACALD"       => -5.65944e-9
  "PTAr"        => 3.41401e-9
  "ALCD2x"      => -1.4328e-9
  "PDH"         => 9.28253
  "PYK"         => 1.75818
  "CO2t"        => -22.8098
  "EX_nh4_e"    => -4.76532
  "MALt2_2"     => -0.0
  "CS"          => 6.00725
  "PGM"         => -14.7161
  "TKT1"        => 1.49698
  "EX_mal__L_e" => -0.0
  "ACONTa"      => 6.00725
  "EX_pi_e"     => -3.2149
  "GLNS"        => 0.223462
  ⋮             => ⋮
```
to this (only the "large" exchange fluxes are shown):
```julia
julia> sol = flux_balance_analysis_dict(model, Tulip.Optimizer)
Consuming fluxes:
        EX_o2_e = -21.799493
        EX_glc__D_e = -10.0
        EX_nh4_e = -4.765319
        EX_pi_e = -3.214895
        ⋮        => ⋮
Producing fluxes:
        EX_h2o_e = 29.175827
        EX_co2_e = 22.809833
        EX_h_e = 17.530865
        ⋮        => ⋮
```
Here `Base.show` is overridden for `Dict{String, Float}` and the exchange fluxes are guessed. I think this display is nicer because you get a sense of the most important attributes from your flux solution immediately. This is also more like what you get from cobrapy 🐍. 

What do you think? If you like it I will wait until #297 is merged because it has some constants that will be used here too. Also, I _really_ want to add colors here... E.g. reaction IDs are one color and fluxes a different color 🎨 @exaexa @laurentheirendt

### DIFF
--- a/docs/src/notebooks/5_basic_stdmodel_construction.jl
+++ b/docs/src/notebooks/5_basic_stdmodel_construction.jl
@@ -27,14 +27,14 @@ gene_list = [Gene(string("g", num)) for num = 1:8]
 #md #       `reactions`. However, these names conflict with generic accessors
 #md #       functions and will create problems downstream.
 
-add!(model, gene_list)
+add_genes!(model, gene_list)
 
 # ### Add metabolites to the model
 metabolite_list = [Metabolite(string("m", num)) for num = 1:4]
 
 metabolite_list[1].formula = "C6H12O6" # can edit metabolites, etc. directly
 
-add!(model, metabolite_list)
+add_metabolites!(model, metabolite_list)
 
 # ### Add reactions to the model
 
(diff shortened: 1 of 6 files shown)


stelmo:
Okay, overloading `show` for `Dict{String, Float}` is perhaps too invasive as @exaexa suggested on slack... Then the question is, do we want to make a new type to display solutions of e.g. `flux_balance_analysis_dict` in a nice way or not?

PR #301, opened 2021-06-06
🩹 fix docs

stelmo:
#299 removed `exchange_reactions` but the tests did not catch that this function was used in the docs - hence master failing... Fix this.


laurentheirendt:
thanks 👍🏼

PR #303, opened 2021-06-07
warmup jump+tulip code to save much time later in testing

### LINKED ISSUE #287: solver precompilation in tests blocks unnecessarily
JuMP+Tulip JIT takes quite a bit of time (~50 secs here). In the tests, the main worker JITs this in `test/base/solver.jl`, the other workers start in FVA tests. That wastest 50 seconds because the JIT doesn't run in parallel althought it could.

*Proposed solution*

kick the solver tests off with something parallel, so that the precompilation is also parallel.

*Outcome*

hopefully -60s test time, much less skewed test timing.

exaexa:
mainly, this should prevent a situation where worker 1 and worker 2 do the
precompilation sequentially, thus wasting 2*N time instead of just N. Actually,
N is now around 50 seconds, so it saves quite a bit of time. Also, time
measurements are now much less skewed by precompilation.

Closes #287

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -17,11 +17,13 @@ using Tulip
 # error tolerance in computations)
 TEST_TOLERANCE = 10 * COBREXA._constants.tolerance
 
+print_timing(fn, t) = @info "$(fn) done in $(round(t; digits = 2))s"
+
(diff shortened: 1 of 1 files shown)


stelmo (approved):
⚡

PR #312, opened 2021-06-08
🩹 ✨ Atoms in reactions and overall flux

### LINKED ISSUE #311: Feature to get atomic counts in reactions
For purposes of flux normalization over carbon-moles I would like to access the amount of carbons within the biomass reaction. Unfortunately there don't seem to be any function to directly access it.
I would suggest a function that takes the atomic count of each participant of the reaction and multiply it with the corresponding stoichiometric value, and add up the resulting numbers for each element. The output could be a dictionary with the elements for keys and the total count within the reaction for values. 
I would be grateful if you consider my suggestion to be included into your module.

stelmo:
Add a function to get the atoms exchanged by a reaction (should be zero for all reactions except the biomass and boundary reactions). See #311 

Also fix issue with `atom_exchange` when applied to overall flux distribution.

Closes #311 when merged.

### DIFF
--- a/src/base/utils/StandardModel.jl
+++ b/src/base/utils/StandardModel.jl
@@ -68,6 +68,22 @@ function atom_exchange(flux_dict::Dict{String,Float64}, model::StandardModel)
 end
 
 """
+    atom_exchange(rxn_id::String, model::StandardModel)
+
+Return a dictionary mapping the flux of atoms through a reaction in `model`.
+"""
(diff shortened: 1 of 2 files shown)


exaexa (approved):
looks good to me
EDIT: except for the tests failing

stelmo:
> looks good to me
> EDIT: except for the tests failing

How important are those really?

PR #315, opened 2021-06-08
hard format check

exaexa:


.gitlab-ci.yml
```diff
@@ -61,17 +61,12 @@ variables:
 # --------------------------------------
 
 format:
-  stage: formatter
+  stage: format-check
   image: $CI_REGISTRY/r3/docker/julia-custom
   script:
     - julia --project=@. -e 'using JuliaFormatter; format(".", verbose=true);'
     - git status
-    - git remote set-url origin "https://$CI_USER_NAME:$GITHUB_ACCESS_TOKEN@github.com/$GITHUB_REPOSITORY"
```
laurentheirendt, line 69: why remove this auto-formatting?

.gitlab-ci.yml
```diff
@@ -1,7 +1,7 @@
 stages:
-  - formatter
   - generator
   - test-containers
+  - format-check
```
laurentheirendt, line 4: I thought having the format-check first would be best before running any tests ...

laurentheirendt (changes requested):

.gitlab-ci.yml
exaexa: It is cool but for whatever reason broken in this form and I can't fix it, so I just axed it.

In the long term I'd kinda like to push everyone to always think about formatting himself and making sure that it looks good before typing `push`. :D

.gitlab-ci.yml
exaexa: if there's no format fixing, this is is better than reporting "hey the code ain't even formatted, I'm not gonna even look at it".

stelmo (approved):
🤖

.gitlab-ci.yml
laurentheirendt: yeah, but that won't happen - I want to keep the auto-formatting. I will look into it to make sure the push works correctly.

.gitlab-ci.yml
laurentheirendt: but the idea is to test with the formatting applied, no? at least that's how I understand it

exaexa: yeah, with autoformatting it didn't make sense to run the tests on unformatted code because they would be run on the formatted either. Without it, it doesn't matter, and it's better to see whether tests fail than whether formatting is wrong.

.gitlab-ci.yml
exaexa: why would it not happen? :D

laurentheirendt: imo, formatting code shouldn't be a requirement before pushing code, and I'd rather rely on auto-formatting to make sure that the PRs are formatted properly before they get merged

.gitlab-ci.yml
laurentheirendt: but then don't we have to test as well the formatted code with the same base image?

laurentheirendt:
I think we can close this - will be revisited in #322

laurentheirendt:
@exaexa can you only put in the formatting? maybe cherry-pick it over

PR #325, opened 2021-06-10
📝 🚧 Community model construction

stelmo:
A short tutorial demonstrating community model construction


docs/src/notebooks/7_community_model.jl
```diff
@@ -0,0 +1,125 @@
...
+using COBREXA
+using Tulip
+
+# ## Load the models and inspect fba solutions
+
+base_model = load_model(CoreModel, "e_coli_core.json")
+
+cytbd_knockout = remove_reactions(base_model, "CYTBD")
+sol = flux_balance_analysis_dict(cytbd_knockout, Tulip.Optimizer)
+println("Cytochrome oxidase knockout μ = ", sol["BIOMASS_Ecoli_core_w_GAM"])
```
exaexa, line 24: println not needed, it's a notebook so you can just let it print itself in one line.

Also, I'd add some coment about what the results actually mean (although it can be seen from the code, the new users will not be able to read it this way)

exaexa: extra suggestion, perhaps rename to eno_knockout_model (very hungarian but helps comprehesion)

docs/src/notebooks/7_community_model.jl
```diff
@@ -0,0 +1,125 @@
...
+println("ATP synthase knockout μ = ", sol["BIOMASS_Ecoli_core_w_GAM"])
+
+eno_knockout = remove_reactions(base_model, "ENO")
+sol = flux_balance_analysis_dict(eno_knockout, Tulip.Optimizer)
+println("Enolase knockout μ = ", sol["BIOMASS_Ecoli_core_w_GAM"])
+# notice that the enolase mutant cannot grow by itself.
+
+# ## Build a community model of the Cytochrome oxidase and ATP synthase knockouts 
+
+ex_rxns = filter(looks_like_exchange_reaction, reactions(base_model))
```
exaexa, line 37: there's a shortcut function for this now

docs/src/notebooks/7_community_model.jl
```diff
@@ -0,0 +1,125 @@
...
+ex_mets = [first(keys(reaction_stoichiometry(base_model, ex_rxn))) for ex_rxn in ex_rxns]
```
exaexa, line 38: needs a comment -- "Extract the metabolite indexes that correspond to the exchange reactions"

docs/src/notebooks/7_community_model.jl
```diff
@@ -0,0 +1,125 @@
...
+
+models = [cytbd_knockout, atps4r_knockout]
```
exaexa, line 40: unnecessary variable, just put it in the function

exaexa: (btw there are 3 knockouts, why just add 2?)

docs/src/notebooks/7_community_model.jl
```diff
@@ -0,0 +1,125 @@
...
+community_model = join_with_exchanges(models, ex_rxns, ex_mets; 
+        add_biomass_objective=true, 
+        biomass_ids=["BIOMASS_Ecoli_core_w_GAM", "BIOMASS_Ecoli_core_w_GAM"], 
+        model_names=["cytbd_ko", "atps4r_ko"]
+        )     
+
+# ## Set bounds of community model
```
exaexa, line 47: perhaps something like "Limit the exchange fluxes" ?

docs/src/notebooks/7_community_model.jl
```diff
@@ -0,0 +1,125 @@
...
+
+for (env_ex, m2_ex, m1_ex) in zip(env_ex_rxn_idxs, cytbd_ex_rxn_idxs, atps4r_ex_rxn_idxs)
+    # in case some exchange reactions are not present in both models
+    m2lb = isnothing(m2_ex) ? 0.0 : atps4r_knockout.xl[m2_ex]
+    m2ub = isnothing(m2_ex) ? 0.0 : atps4r_knockout.xu[m2_ex]
+
+    m1lb = isnothing(m1_ex) ? 0.0 : cytbd_knockout.xl[m1_ex]
+    m1ub = isnothing(m1_ex) ? 0.0 : cytbd_knockout.xu[m1_ex]
+
+    change_bounds!(community_model, [env_ex]; xl= [m1lb + m2lb], xu = [m1ub + m2ub])
```
exaexa, line 61: side note, there's `set_bound` function, this seems to do the same, right?

Can we rename it later?

docs/src/notebooks/7_community_model.jl
```diff
@@ -0,0 +1,125 @@
...
+end
+
+# ## Add objective function to community model
+
+biomass_ids = ["cytbd_ko_BIOMASS_Ecoli_core_w_GAM", "atps4r_ko_BIOMASS_Ecoli_core_w_GAM"]
```
exaexa, line 66: if you had model_names in a variable, you could do
`model_names .* "_BIOMASS_Ecoli_core_w_GAM" `

docs/src/notebooks/7_community_model.jl
```diff
@@ -0,0 +1,125 @@
...
+add_objective!(
```
exaexa, line 67: naming consistency: this should be called add_ or set_ ?

(perhaps add_to_objective?)

docs/src/notebooks/7_community_model.jl
```diff
@@ -0,0 +1,125 @@
...
+    community_model,
+    biomass_ids;
+    objective_column_index = first(
+        indexin(["community_biomass"], reactions(community_model)),
+    ),
+)
+
+# ## Perform community FBA
```
exaexa, line 75: this sounds like there's a special cFBA  :D "Run FBA on the community" ?

docs/src/notebooks/7_community_model.jl
```diff
@@ -0,0 +1,125 @@
...
+
+d = flux_balance_analysis_dict(
+    community_model,
+    Tulip.Optimizer;
+    modifications = [change_optimizer_attribute("IPM_IterationsLimit", 1000)],
+)
+println("Community μ = ", d["community_biomass"])
```
exaexa, line 82: better do comment + variable display, not explicit printing

docs/src/notebooks/7_community_model.jl
exaexa: EDIT: I found out why you start with 2 only. :D

docs/src/notebooks/7_community_model.jl
```diff
@@ -0,0 +1,125 @@
...
+community_model = add_model_with_exchanges(
+    community_model,
+    eno_knockout,
+    ex_rxns,
+    ex_mets;
+    model_name = "eno_ko",
+    biomass_id = "BIOMASS_Ecoli_core_w_GAM",
+)
+
+# ## update the biomass reaction to include the new mutant
```
exaexa, line 96: this may be a bit misleading, it doesn't update the reaction but the objective right?

exaexa: (also another heading here may be just superfluous)

docs/src/notebooks/7_community_model.jl
```diff
@@ -0,0 +1,125 @@
...
+    ),
+)
+
+d = flux_balance_analysis_dict(
+    community_model,
+    Tulip.Optimizer;
+    modifications = [change_optimizer_attribute("IPM_IterationsLimit", 1000)],
+)
+println("Community μ = ", d["community_biomass"])   
+# Notice that the high communal growth rate is 0, due to the enolase knockout.
```
exaexa, line 113: println → implicit print with comment

exaexa: anyway this might deserve a bit of description for people who hear about enolase the first time in their life

docs/src/notebooks/7_community_model.jl
```diff
@@ -0,0 +1,125 @@
...
+
+# ## Allow the mutants to share pyruvate and rescue each other
```
exaexa, line 115: Cool! maybe reverse the order of actions "to rescue each other by sharing pyruvate"

docs/src/notebooks/7_community_model.jl
```diff
@@ -0,0 +1,125 @@
...
+
+pyr_exs = ["cytbd_ko_EX_pyr_e", "atps4r_ko_EX_pyr_e", "eno_ko_EX_pyr_e"]
+change_bounds!(community_model, pyr_exs; xl = repeat([-1000.0], inner=3), xu = repeat([1000.0], inner=3))
```
exaexa, line 118: better use fill(-1000.0, 3)?

docs/src/notebooks/7_community_model.jl
exaexa: semantics question, why is the community biomass production 0 when the other 2 mutants should be able to surive themselves? (and, of course, feast on the dead cytoplasm of the third one :zombie: )

docs/src/notebooks/7_community_model.jl
```diff
@@ -0,0 +1,125 @@
...
+
+d = flux_balance_analysis_dict(
+    community_model,
+    Tulip.Optimizer;
+    modifications = [change_optimizer_attribute("IPM_IterationsLimit", 1000)],
+)
+println("Community μ = ", d["community_biomass"])
```
exaexa, line 125: println → comment + implicit print

exaexa (changes requested):
some wording changes

docs/src/notebooks/7_community_model.jl
stelmo: yeah, but I like the filter approach more, it introduced the `looks_like...` etc. functions :)

docs/src/notebooks/7_community_model.jl
stelmo: `Set exchange reaction bounds of community model based on the bounds of the individual models`

docs/src/notebooks/7_community_model.jl
stelmo: now we do :)

docs/src/notebooks/7_community_model.jl
stelmo: it actually does both adding and setting... Adding if it is not present and setting if it is :/ A two for one function

docs/src/notebooks/7_community_model.jl
stelmo: lol, there is actually and this is it :)

docs/src/notebooks/7_community_model.jl
stelmo: removed the heading but the objective is also a physical reaction in the model

exaexa (approved):

PR #327, opened 2021-06-10
🩹 Make `atom_exchage` robust to metabolites missing formulas

### LINKED ISSUE #326: 🩹 `atom_exchange` is not robust to missing formulas
`atom_exchange` breaks when applied to metabolites without formulas - fix this to prevent scripts breaking mid run.

Maybe something like this, although `try..catch` may have a performance penalty?
```julia
function atom_exchange2(flux_dict::Dict{String,Float64}, model::StandardModel)
    atom_flux = Dict{String,Float64}()
    for (rxn_id, flux) in flux_dict
        if is_boundary(model.reactions[rxn_id])
            for (met, stoich_rxn) in model.reactions[rxn_id].metabolites
                try 
                    adict = get_atoms(model.metabolites[met])
                    for (atom, stoich_molecule) in adict
                        atom_flux[atom] = get(atom_flux, atom, 0.0) + flux * stoich_rxn * stoich_molecule
                    end
                catch err
                    # println("error getting atoms of: ", met, " in reaction: ", rxn_id)
                    continue
                end
            end
        end
    end
    return atom_flux
end
end

### PR DESCRIPTION
**Pull request by stelmo**
Addresses #326 by adding a continue clause for `atom_exchange`

Closes #326 when merged.

### DIFF
--- a/src/base/utils/Reaction.jl
+++ b/src/base/utils/Reaction.jl
@@ -61,7 +61,7 @@ function is_mass_balanced(rxn::Reaction, model::StandardModel)
     atom_balances = Dict{String,Float64}() # float here because stoichiometry is not Int
     for (met, stoich) in rxn.metabolites
         atoms = metabolite_formula(model, met)
-        isempty(atoms) && continue # ignore blanks
+        isnothing(atoms) && continue # ignore blanks
         for (k, v) in atoms
             atom_balances[k] = get(atom_balances, k, 0) + v * stoich
         end
--- a/src/base/utils/Reaction.jl
+++ b/src/base/utils/Reaction.jl
@@ -60,7 +60,7 @@ See also: [`get_atoms`](@ref), [`check_duplicate_reaction`](@ref)
 function is_mass_balanced(rxn::Reaction, model::StandardModel)
     atom_balances = Dict{String,Float64}() # float here because stoichiometry is not Int
     for (met, stoich) in rxn.metabolites
-        atoms = get_atoms(model.metabolites[met])
+        atoms = metabolite_formula(model, met)
         isempty(atoms) && continue # ignore blanks
         for (k, v) in atoms
             atom_balances[k] = get(atom_balances, k, 0) + v * stoich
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/base/utils/Reaction.jl [hunk 1]
```diff
@@ -60,7 +60,7 @@ See also: [`get_atoms`](@ref), [`check_duplicate_reaction`](@ref)
 function is_mass_balanced(rxn::Reaction, model::StandardModel)
     atom_balances = Dict{String,Float64}() # float here because stoichiometry is not Int
     for (met, stoich) in rxn.metabolites
-        atoms = get_atoms(model.metabolites[met])
+        atoms = metabolite_formula(model, met)
         isempty(atoms) && continue # ignore blanks
```
exaexa, line 64: `isnothing` here

(there's nothing wrong with it being empty, just an empty molecule, carry on nothing to see :D )

src/base/utils/StandardModel.jl
```diff
@@ -76,7 +77,8 @@ Return a dictionary mapping the flux of atoms through a reaction in `model`.
 function atom_exchange(rxn_id::String, model::StandardModel)
     atom_flux = Dict{String,Float64}()
     for (met, stoich_rxn) in model.reactions[rxn_id].metabolites
-        adict = get_atoms(model.metabolites[met])
+        adict = metabolite_formula(model, met)
+        isnothing(adict) && continue
```
exaexa, line 81: I so much wish there was a function for this like with haskell, this is double `traverse`, and you don't need to care about whether it's nothing or empty or whatever. :D

exaexa (approved):
pls switch one isempty to isnothing (`nothing` is magically marked empty but I don't understand why, it carries one negative information right? :D )

src/base/utils/Reaction.jl
stelmo: wow, this was done too late at night... I thought I used isnothing :/

stelmo:
> EDIT please triplecheck formatting

what do you mean "EDIT please triplecheck formatting"?

PR #331, opened 2021-06-11
make atom_exchange and metabolite_fluxes a bit more generic

### LINKED ISSUE #317: generalize `get_atom_exchanges`
- The current implementation only works with a StandardModel (there's no reason for that, all required information is available generically)
- we could have nice `atom_stoichiometry` that returns a stoichiometry matrix where metabolites are compartmentalized atoms

exaexa:
Closes #317


stelmo (approved):
This will break master because I am pretty sure some of the utility functions with the order changed are documented and the docs have not been updated :)

exaexa:
I grepped the whole thing and fixed the docs accordingly, so hopefully not. At least the docs built here OK.

Issue #338, opened 2021-06-13
💄 📝 🚸 Small improvements


stelmo:
Some functions have very spartan documentation. List them here for improvement (🚑 = important):

- [ ] `add_reactions` 

Some functions need to be changed a little bit:

- [x] `Reaction` let reaction equation dicts also do `Dict{String, Int}` as opposed to `Dict{String, Float}` only
- [x] `change_bounds!` also doesn't accept Ints and gives a terrible error message
- [x] `add_objective!` - returns the last assigned value, which looks weird, make it return nothing
- [x] `flux_summary` - prints unbounded reactions
- [x] `load_model(StandardModel,...)` the conversion is slow, make it faster
- [x] `save_json_model` is painfully slow - the file shows up but is empty for 10s of seconds before things actually get written to it... it makes you think something has gone wrong. 🚑 
- [x] `is_mass_balanced` should issue a warning if a metabolite does not have a formula and you are checking if a reaction is mass balanced - false negatives could be generated...
- [ ] `remove_reaction(!)` is in place for `StandardModel` and returns a new model for `CoreModel` this makes the function signature slightly irritating to use since for the one model type you need a bang and for the other you don't.. maybe add in place versions for the `CoreModel` functions? 

Note: I will add more as I use COBREXA - don't address this in a PR yet!

stelmo:
Hmm, `change_bounds!` does not work as expected... It seems to overwrite more than just the supplied reaction bounds... 
```julia
using COBREXA
using Gurobi

model_path = joinpath("models", "e_coli_core.json")
model = load_model(StandardModel, model_path)

original_ex_rxns = filter(looks_like_exchange_reaction, reactions(model))
non_ex_mets = setdiff(metabolites(model), filter(looks_like_exchange_metabolite, metabolites(model)))

rvec = [Reaction("EX_$met", Dict(met => -1.0), :bidirectional) for met in non_ex_mets]
add_reactions!(model, rvec)

minimal_media = Dict("EX_o2_e" => (-20.0, 1000.0), 
                     "EX_glc__D_e" => (-1000.0, 1000.0), 
                     "EX_nh4_e" => (-1000.0, 1000.0), 
                     "EX_pi_e" => (-1000.0, 1000.0),
                     "EX_h_e" => (-1000.0, 1000.0), 
                     "EX_co2_e" => (-1000.0, 1000.0), 
                     "EX_h2o_e" => (-1000.0, 1000.0))

[...]

PR #342, opened 2021-06-17
don't leak the file object when saving JSON

exaexa:
refs :lipstick: #338


laurentheirendt (approved):

stelmo (approved):
why have 3 lines when you can do it in one :)

PR #384, opened 2021-07-10
some cosmetic README touches, dev doc links

exaexa:
@laurentheirendt we could also have versioned docs archived now right?


cylon-x:
:green_circle: &nbsp;Commit e12e1712 is formatted properly.

laurentheirendt (approved):

PR #395, opened 2021-07-26
improve jsonmodel speed + some extra cleanup

exaexa:
Fixes #350 by caching.

PR #396, opened 2021-07-27
Copy of #395 to test the tests

exaexa:
cc @laurentheirendt


exaexa:
OK this was naive, sorry for the noise.

Issue #382, opened 2021-07-10
Add latex completion hints for non-ascii characters in docs


stelmo:
Look at the docstring of ⟶ and → in the docs, they look identical. This could be confusing, add the latex shortcuts to the docs to make the difference obvious, e.g. in this case `\rightarrow<tab> = →` and `\longrightarrow<tab> = ⟶`

exaexa:
I still kinda think that we shold avoid having both → and →.

stelmo:
ah okay, I am open to deleting one of them actually! Maybe the shorter one?

exaexa:
keep →, delete the long variant (it would be weird to only have an adjectivized variant)

stelmo:
but but but the long one looks prettier :)

stelmo:
The more I read my own comments, the more I realize that I just love good looking code/docs

exaexa:
I wish we could just write the nice interoperable `->` and have editors that display it the right way.

stelmo:
So it turns out I already added the latex completion stuff to the docs! https://lcsb-biocore.github.io/COBREXA.jl/stable/notebooks/5_basic_stdmodel_construction/#Add-reactions-to-the-model

stelmo:
hmm although these arrows kinda clash with ligatures... Like it is hard to distinguish -> from \longrightarrow

PR #402, opened 2021-07-28
simplify code by removing unnecessary unicode

exaexa:
- unify arrows to a single form →
- remove lots of other unicode from code

### DIFF
--- a/docs/src/notebooks/5_basic_stdmodel_construction.jl
+++ b/docs/src/notebooks/5_basic_stdmodel_construction.jl
@@ -63,6 +63,15 @@ end
 
 model.reactions["r4"].grr = [["g5"], ["g6", "g7"], ["g8"]]
 
+#md # !!! note "Note: Writing unicode arrows"
+#md #     The reaction arrows can be easily written by using the `LaTeX`
+#md #     completions built into Julia shell (and many Julia-compatible
+#md #     editors). You can type:
+#md #
+#md #     - `→` as `\rightarrow` (press `Tab` to complete)
+#md #     - `←` as `\leftarrow`
+#md #     - `↔` as `\leftrightarrow`
+
 # The constructed model can now be inspected.
 model
 
--- a/docs/src/notebooks/5_basic_stdmodel_construction.jl
+++ b/docs/src/notebooks/5_basic_stdmodel_construction.jl
@@ -41,7 +41,7 @@ add_metabolites!(model, metabolite_list)
 # There are two ways to create and add reactions to a model.
 # These are using functions, or macros.
 
-r_m1 = Reaction("EX_m1", Dict("m1" => -1.0), :bidirectional) # exchange reaction: m1 <-> (is the same as m1 ⟷ nothing)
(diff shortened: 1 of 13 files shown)


exaexa:
( Kinda removes the need to solve #382 )

cylon-x:
:green_circle: &nbsp;Commit 5c37fe4f is formatted properly.

docs/src/notebooks/5_basic_stdmodel_construction.jl
```diff
@@ -55,28 +55,14 @@ m3 = metabolite_list[3]
...
 
 model.reactions["r4"].grr = [["g5"], ["g6", "g7"], ["g8"]]
 
-#md # !!! note "Note: Using reaction arrows"
-#md #       `COBREXA` exports arrows that can be used to construct reactions.
-#md #       Both the long and short arrows (`⟶ == →`) mean the same thing if they
-#md #       point in the same direction.
-#md #
-#md #       These arrows are accessible by using the `LaTeX` completions built into
-#md #       Julia. For example:
```
laurentheirendt, line 72: maybe it is good to keep a note on how to enter these arrows 😄

laurentheirendt: I actually think that the long version of these arrows provides an increased readability, not sure what you are thinking of it.

laurentheirendt (changes requested):

docs/src/notebooks/5_basic_stdmodel_construction.jl
stelmo: Yep, I also like the long versions more, they display better. Hmm now that I have ligatures, it would be cool if we could replace the arrows with <-->, <--, and -->

stelmo: Is it even possible (easy) to do that? How does julia interpret things like `-->`, can we define what that means somehow? It would be really cool

stelmo (approved):

docs/src/notebooks/5_basic_stdmodel_construction.jl
exaexa: custom operators are not supported by julia

\bigfrown

cylon-x:
:green_circle: &nbsp;Commit 8a85a78c is formatted properly.

docs/src/notebooks/5_basic_stdmodel_construction.jl
exaexa: (I added the note back)

laurentheirendt (approved):

laurentheirendt:
shall we then leave the short arrows or shall we go for the long ones?

exaexa:
> shall we then leave the short arrows or shall we go for the long ones?

I wrote a comment here arguing for the short ones but it got somehow erased (githubs..)

- short render basically the same as long ones in terminals, the double width somehow doesn't work for long ones
- if tabbing a latex command, most people dont start search for right arrow by typing "long"
- short variants are properly supported by all compose systems I know (compose + `-` + `>` writes the "normal" ie short one)

PR #449, opened 2021-09-07
correctly forward the `workers` parameter

exaexa:
## Description

Amusing.

I wish *this* was the cause of the benchmark slowness. :D


cylon-x:
:green_circle: &nbsp;Commit d08bff84 is formatted properly.

PR #450, opened 2021-09-09
completely disable JuMP bridge constraints

exaexa:
We do not use them at all, which is likely gonna hold for some time. This also speeds up the precompilation somewhat, and removes some unneeded overhead from the model logic.


cylon-x:
:green_circle: &nbsp;Commit e99747f6 is formatted properly.

laurentheirendt (approved):

PR #452, opened 2021-09-09
fix the index in remove_metabolites of CoreModel

exaexa:


cylon-x:
:green_circle: &nbsp;Commit b4e8a9e9 is formatted properly.

stelmo (approved):
Whoops 😅

src/reconstruction/CoreModel.jl
```diff
@@ -402,7 +402,7 @@ end
         model,
         [
             ridx for ridx in 1:n_reactions(model) if
-            any(in.(findnz(model.S[:, ridx])[2], Ref(metabolite_idxs)))
+            any(in.(findnz(model.S[:, ridx])[1], Ref(metabolite_idxs)))
```
htpusa, line 405: What exactly is this function doing?

exaexa: Checking which reactions are to be removed (because they touch the metabolite).

htpusa: Oh it removes any reaction that has a metabolite being removed, of course. Is that the default assumption (that one would want all reactions affected to be gone)?

exaexa: yeah. after some thinking we assumed this way to be better, because it prevents breaking semantics of single reactions (e.g. if you have "transport of something" and remove 1 of the 2 metabolites, you suddenly get "infinite supply of something"). This way is more "conservative".

PR #454, opened 2021-09-15
Revert "completely disable JuMP bridge constraints"

exaexa:
This reverts commit e99747f6ca279ffe3b5d9e691af76fc79cd90ad4.

Apparently some internal JuMP logic depends on the bridges, preventing e.g.
single-variable objective functions from working with GLPK (and maybe other
solvers). Dodging that is easy, e.g. you can use `@objective(m, Max, x+0)`
instead of one that contains just `x`, but kludges like this should not be
required in the code.


cylon-x:
:green_circle: &nbsp;Commit 0cb0e452 is formatted properly.

laurentheirendt (approved):

PR #483, opened 2021-11-02
Simplify the develop CI

exaexa:


cylon-x:
:green_circle: &nbsp;Commit 2ca7d725 is formatted properly.

.gitlab-ci.yml
```diff
@@ -108,7 +108,7 @@ format:
 # Documentation history
 # --------------------------------------
 
-generator:gource:
+gource:
```
laurentheirendt, line 111: why?

docs/src/index.md.template
```diff
@@ -32,7 +32,7 @@ the popular [`Tulip.jl`](https://github.com/ds4dm/Tulip.jl),
 
 ```@raw html
 <div align="center">
-<img style="width:300px;margin:10px;border-offset:15px;border: 1px solid #eee;border-radius: 50%;padding: 10px;-webkit-border-radius: 50%;-moz-border-radius: 50%;" src="https://git-r3lab.uni.lu/lcsb-biocore/COBREXA.jl/-/jobs/artifacts/master/raw/output.gif?job=generator:gource" alt="history"><br>
+<img style="width:300px;margin:10px;border-offset:15px;border: 1px solid #eee;border-radius: 20%;padding: 10px;" src="https://git-r3lab.uni.lu/lcsb-biocore/COBREXA.jl/-/jobs/artifacts/master/raw/output.gif?job=gource" alt="history"><br>
```
laurentheirendt, line 35: how does it look like? Can you also change the gitlab url?

.gitlab-ci.yml
```diff
@@ -140,8 +140,6 @@ doc-tests:v1.6:
 
 pages:
   stage: documentation
-  dependencies:
```
laurentheirendt, line 140: I think it is still required, no? I think we cannot assume that gource is always running ...

laurentheirendt (changes requested):

.gitlab-ci.yml
exaexa: the jobs are technically independent, and I wanted the gource to be built only from "released" stuff on master (stuff in develop might be messy/reorganized).

.gitlab-ci.yml
exaexa: I didn't get the semantics of the generator: prefix (there's none I guess?) so I just removed it.

docs/src/index.md.template
exaexa: It's IMO better (doesn't kill 30% of the nice picture area, just like 5%):
image

The URL got changed because the step name changed. (I hope the prefix wasn't there just because of this.)

.gitlab-ci.yml
laurentheirendt: no there is 😄 it is a substage of `generator`

docs/src/index.md.template
laurentheirendt: hmmm, the idea was to keep a circle, like a lens

.gitlab-ci.yml
laurentheirendt: the jobs are not really independent - like the deployment of the docs should not happen if gource has failed

.gitlab-ci.yml
exaexa: but there is no generator anymore? :D

docs/src/index.md.template
exaexa: ah so. I kinda imagined a submarine window so far.... :D

.gitlab-ci.yml
exaexa: ah okay, that makes sense

.gitlab-ci.yml
laurentheirendt: there used to be I think - I'd suggest to leave it to easily extend later on if necessary

docs/src/index.md.template
laurentheirendt: not sure why we have to change :D

exaexa:
ok let's not do this yet, current pipeline hopefully works, that's it. :D

PR #495, opened 2021-11-03
develop→master

exaexa:


cylon-x:
:red_square: &nbsp;Commit 44d64482 requires formatting!

Required formatting changes summary:
```
```

stelmo (approved):
🚀

docs/make.jl
```diff
@@ -2,16 +2,11 @@ using Documenter
...
-if "CI_EXTERNAL_PULL_REQUEST_IID" in keys(ENV)
-    ENV["TRAVIS_PULL_REQUEST"] = ENV["CI_EXTERNAL_PULL_REQUEST_IID"]
-else
-    ENV["TRAVIS_PULL_REQUEST"] = "false"
-end
+# some settings
+dev_docs_folder = "dev"
+pages_branch = "gh-pages"
+github_repo_slug = "LCSB-BioCore/COBREXA.jl"
+delete!(ENV, "GITHUB_REPOSITORY")
```
laurentheirendt, line 9: I think we should remove this env variable on Gitlab and not here - can you change?

docs/make.jl
```diff
@@ -2,16 +2,11 @@ using Documenter
...
-# set the merge/pull request ID
-if "CI_EXTERNAL_PULL_REQUEST_IID" in keys(ENV)
-    ENV["TRAVIS_PULL_REQUEST"] = ENV["CI_EXTERNAL_PULL_REQUEST_IID"]
-else
-    ENV["TRAVIS_PULL_REQUEST"] = "false"
-end
+# some settings
+dev_docs_folder = "dev"
+pages_branch = "gh-pages"
+github_repo_slug = "LCSB-BioCore/COBREXA.jl"
```
laurentheirendt, line 8: this should be retrievable from the env

README.md
```diff
@@ -18,7 +18,7 @@
 [cov-url]: https://codecov.io/gh/LCSB-BioCore/COBREXA.jl
 
 [contrib-img]: https://img.shields.io/badge/contributions-start%20here-green
-[contrib-url]: https://github.com/LCSB-BioCore/COBREXA.jl/blob/master/.github/CONTRIBUTING.md
+[contrib-url]: https://lcsb-biocore.github.io/COBREXA.jl/dev/howToContribute/
```
laurentheirendt, line 21: The URL should be `stable` here I think

exaexa: contribution kindof implies development so I pointed it to the (likely more relevant) dev version

docs/make.jl
exaexa: I can't modify gitlab settings but you can safely erase this one there now. Then I can remove this from here

docs/make.jl
exaexa: Doesn't seem so, I'd need to parse it out of EXTERNAL_REPO.

README.md
laurentheirendt: contributing guide should be the one from `master`, i.e. the stable version. 

Although a developer might consult the development guide, it should be the stable version

docs/make.jl
laurentheirendt: this script executes on gitlab CI, so we can use `CI_PROJECT_NAMESPACE` and `CI_PROJECT_NAME`

docs/make.jl
laurentheirendt: we can remove it, but then let's use other variables (see comment below)

README.md
exaexa: ok

docs/make.jl
exaexa: but that doesn't match the github ones (that I need) right?

laurentheirendt: it is the same as the gitlab one, as it is a mirror

exaexa: ok, fixed in #500

docs/make.jl
exaexa: Removed completely (see #500)

Issue #474, opened 2021-10-27
Unify the `_dict` and `_vec` function variants with some common reusable code


exaexa:
For example the `flux_balance_analysis_vec` and now `minimize_metabolic_adjustment_vec` shouldn't reimplement the same thing.

PR #514, opened 2021-11-11
simplify the _dict and _vec variants of analysis functions

exaexa:
...using some shared code.

In turn I believe the change is not breaking, the functions are now just equipped with kinda deprecation warnings.

Also this bumps the version as #474 was the last issue for the milestone.


cylon-x:
:green_circle: &nbsp;Commit 0079d0b8 is formatted properly.

stelmo (approved):
nice :)

PR #527, opened 2021-11-26
make sure SBML 0.8.1 doesn't get installed as a dep

exaexa:
(it doesn't work because of the units problem there)


cylon-x:
:green_circle: &nbsp;Commit 739c6d45 is formatted properly.

laurentheirendt (approved):

PR #536, opened 2021-12-04
bump up the "history gif" artifact expiration to 3 months

exaexa:
Now that we don't refresh master that often, the default expire rule leaves us
without the history animation most of the time. This should add a bit of the
bumper space.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -221,6 +221,7 @@ documentation-assets:gource:
     - docker run -v "$PWD":/visualization $CI_REGISTRY/r3/docker/gource
   artifacts:
     paths: ['output.gif']
+    expire_in: 3 mos
   <<: *global_trigger_build_doc
 
 #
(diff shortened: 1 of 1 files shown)


cylon-x:
:green_circle: &nbsp;Commit 9d37f2d2 is formatted properly.

cylon-x:
:green_circle: &nbsp;Commit cad9f111 is formatted properly.

.gitlab-ci.yml
```diff
@@ -221,6 +221,7 @@ documentation-assets:gource:
     - docker run -v "$PWD":/visualization $CI_REGISTRY/r3/docker/gource
   artifacts:
     paths: ['output.gif']
+    expire_in: 3 mos
```
laurentheirendt, line 224: I'd increase even to 1 year or longer

laurentheirendt (changes requested):

.gitlab-ci.yml
stelmo: you mean 12 mos

exaexa: I hope that it can parse "year" (the doc is not very authoritative about what it can actually do :D). If not, we go to Mo's.

cylon-x:
:green_circle: &nbsp;Commit 70d7bab2 is formatted properly.

laurentheirendt (approved):

PR #537, opened 2021-12-06
Regular merge develop→master

exaexa:


Project.toml
```diff
@@ -4,7 +4,6 @@ authors = ["The developers of COBREXA.jl"]
 version = "1.1"
```
laurentheirendt, line 4: I think you have to bump the version here

laurentheirendt (changes requested):

test/data_static.jl
```diff
@@ -72,3 +72,79 @@ test_toyModel() = CoreModel(
     ["r1", "m1t", "m3t", "EX_m1(e)", "EX_m3(e)", "EX_biomass(e)", "biomass1"],
     ["m1[c]", "m3[c]", "m2[c]", "m1[e]", "m3[e]", "biomass[c]"],
 )
+
+const reaction_standard_gibbs_free_energies = Dict(
```
laurentheirendt, line 76: how was this data actually generated?

Project.toml
exaexa: Ah I didn't want to version this but we can do, OK

test/data_static.jl
exaexa: cc @stelmo 

I thought I asked for this in review as well, AFAIK it's described in tutorials (where the same data is used) but yeah a link would help here a lot.

Project.toml
laurentheirendt: I think it is always the idea when merging `develop` ->`master` to then tag the repo (even patch)

test/data_static.jl
stelmo: cool, will add one

Project.toml
exaexa: ok

laurentheirendt (approved):

PR #561, opened 2022-01-11
COBREXA should be a dep of documentation Project.toml to enforce version checks

exaexa:


laurentheirendt (approved):
let's depend on ourselves 😄

exaexa:
yeah that's something dependable for sure! :D

Issue #112, opened 2021-04-22
Bounds should not be sparse vectors, maybe a new data type...


stelmo:
Bound vectors are typically not sparse in the usual sense. They are populated with max/min bounds and not very many zeros. It might make sense to make our own "sparse" vector format where the zeros are actually the max or min bounds. This might save significant storage at the exa-scale...

PR #562, opened 2022-01-24
postpone JuMP 0.22 compatibility

exaexa:
This would break most of the tutorials and notebooks before Tulip issue #112 is
addressed (see https://github.com/ds4dm/Tulip.jl/issues/112).


cylon-x:
:green_circle: &nbsp;Commit 9ec87385 is formatted properly.

laurentheirendt (approved):

PR #622, opened 2022-05-16
Fix small error with gapfill modifications

stelmo:
A type checker would've been nice right about now :)


exaexa (approved):
...yeah, typechecker or full test coverage :D

PR #627, opened 2022-05-26
sampling improvements

exaexa:
Still needs a test.

@stelmo would you have a good sampling model with reaction coupling? Do we just throw in one with gecko?


stelmo:
> Still needs a test.
> 
> @stelmo would you have a good sampling model with reaction coupling? Do we just throw in one with gecko?

Hmm I would just do Gecko

PR #628, opened 2022-06-06
docs reorganization

exaexa:
## Description

This should solve some of the documentation issues that got accumulated during the development. Mostly, this focuses on easy availability of code to run that does various things, instead of having super-specific tutorial with long complex analyses.

## Organization

cc @stelmo 

Examples (presented as notebooks):
- [x] Loading models
- [x] Converting and saving models
- [x] Exploring models
- [x] Constructing models
- [x] FBA
- [x] FVA
- [x] Gene deletions
- [x] Single-reaction modifications
- [x] pFBA
- [x] loopless FBA
- [x] Enzyme crowding
- [x] Growth media (the MILP version is omitted now)
- [x] MMDF
- [x] MOMA
- [x] sMOMENT
- [x] GECKO
- [x] Hit&run sampling
- [x] Production envelopes

Parallel and distributed processing:
- [x] Parallelizable functions
- [x] Local parallelism
- [x] Running in Slurm
- [x] Efficient data transfers in HPCs (aka Serialized)

Core concepts:
- [x] Creating and implementing model variants
- [x] Optimizer modifications
- [x] Creating custom models (aka MetabolicModel structure)
- [x] Extending the models (aka ModelWrapper)

Extra things not to forget:
- [x] grep TODO
- [x] `git rm -fr docs/src/notebook*`


stelmo:
also, Types and Functions => is there a way to only display the most generic function, and not each type dispatched variant?

exaexa:
> Don't forget about the nonfunctional quickstart example with the missing "R_" in the readme

well check the commits

> also, Types and Functions => is there a way to only display the most generic function, and not each type dispatched variant?

probably not in general (also, Julia likely doesn't have the concept of "most generic"). Any specific place where this would help? (I guess the modification overloads?)

stelmo:
> probably not in general (also, Julia likely doesn't have the concept of "most generic"). Any specific place where this would help? (I guess the modification overloads?)

Hmm actually, if the examples are done well I don't think this is super necessary (except maybe to group them to make finding the method signatures easier)

stelmo (approved):
Rock and roll!

PR #636, opened 2022-06-21
make reaction_atom_balance have generic MetabolicModel as type

HettieC:
changed reaction_atom_balance to have MetabolicModel as type


exaexa:
Looks good but we have problems with testing from PRs. Will test manually.

HettieC:
Just done juliaformatter, cheers

exaexa (approved):
CI passes locally, formatting is OK just doesn't report from the forks.

PR #637, opened 2022-06-27
remove lots of lb/ub shortcuts from API

### LINKED ISSUE #623: Consistent argument/kwarg naming
Make all the args/kwargs consisten, e.g. `change_constraint(...; lb=..., ub=...)` vs `change_bound(....; lower=..., upper=...)`

exaexa:
Closes #623 

This is breaking, do not merge until we are sure that 2.0 is going to be released next.

We might possibly also remove `grr` and other shortcuts.

### DIFF
--- a/src/base/solver.jl
+++ b/src/base/solver.jl
@@ -73,11 +73,11 @@ not be changed.
 function set_optmodel_bound!(
     vidx,
     opt_model;
-    lb::Maybe{Real} = nothing,
-    ub::Maybe{Real} = nothing,
+    lower::Maybe{Real} = nothing,
+    upper::Maybe{Real} = nothing,
 )
-    isnothing(lb) || set_normalized_rhs(opt_model[:lbs][vidx], -lb)
-    isnothing(ub) || set_normalized_rhs(opt_model[:ubs][vidx], ub)
+    isnothing(lower) || set_normalized_rhs(opt_model[:lbs][vidx], -lower)
+    isnothing(upper) || set_normalized_rhs(opt_model[:ubs][vidx], upper)
 end
 
 """
--- a/test/base/utils/StandardModel.jl
+++ b/test/base/utils/StandardModel.jl
@@ -15,10 +15,10 @@
     glucose_index = first(indexin(["EX_glc__D_e"], reactions(model)))
     o2_index = first(indexin(["EX_o2_e"], reactions(model)))
     atpm_index = first(indexin(["ATPM"], reactions(model)))
-    set_optmodel_bound!(glucose_index, cbm; ub = -1.0, lb = -1.0)
+    set_optmodel_bound!(glucose_index, cbm; upper = -1.0, lower = -1.0)
     @test normalized_rhs(ubs[glucose_index]) == -1.0
     @test normalized_rhs(lbs[glucose_index]) == 1.0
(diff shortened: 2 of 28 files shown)


exaexa:
cc @stelmo :]

exaexa:
Will need to fix the tests but yes it's a part of better organizing all the stuff

src/base/solver.jl
```diff
@@ -69,8 +69,8 @@ get_optmodel_bounds(opt_model) = (
 
 """
     set_optmodel_bound!(vidx, opt_model;
-        ub::Maybe{Real} = nothing,
-        lb::Maybe{Real} = nothing,
+        lower::Maybe{Real} = nothing,
+        upper::Maybe{Real} = nothing,
```
stelmo, line 73: Should all of these be lower_bound and upper_bound as well? I know it is implied by the bound, but then the user needs to remember two kwargs, and not just the one...

test/base/utils/StandardModel.jl
```diff
@@ -15,10 +15,10 @@
     glucose_index = first(indexin(["EX_glc__D_e"], reactions(model)))
     o2_index = first(indexin(["EX_o2_e"], reactions(model)))
     atpm_index = first(indexin(["ATPM"], reactions(model)))
-    set_optmodel_bound!(glucose_index, cbm; ub = -1.0, lb = -1.0)
+    set_optmodel_bound!(glucose_index, cbm; upper = -1.0, lower = -1.0)
```
stelmo, line 18: Same comment about adding `_bound`

src/base/solver.jl
```diff
@@ -80,11 +80,11 @@ not be changed.
 function set_optmodel_bound!(
     vidx,
     opt_model;
-    lb::Maybe{Real} = nothing,
-    ub::Maybe{Real} = nothing,
+    lower::Maybe{Real} = nothing,
```
stelmo, line 83: Same `_bound` comment

stelmo: Maybe check that I did not miss a `lower` or `upper` somewhere

src/base/solver.jl
exaexa: here the `bound` is in the name already so I omitted it. Does it make sense?

Same goes for `change_bound[s][!]`.

stelmo: Hmm but that is too broad, because S*v = 0 is also a constraint

exaexa: yap. Naming is hard. :D

exaexa:
AFTER SOME CONVERSATION, let's postpone this a bit (it's breaking anyway so needs to go to 2.0), and improve even more:
- we can have submodules for modification, reconstruction variants and wrapper constructors to organize them and allow discovering stuff with `[tab]`
- tenative submodules: `mod` for jump modifiers, `change` for reconstruction stuff (no matter if inplace or not), `make` for wrappers/builders, `with` for pipeable reconstruction/wrappers

exaexa:
Closes #623

src/base/solver.jl
exaexa: checked with grep

src/base/solver.jl
exaexa: we'll see after everything gets migrated to new names on how necessary the _bound is after that

PR #644, opened 2022-08-01
extend the autoformatting priveleges a bit

exaexa:
( Let's use this moment to silently, deeply reflect on the github privilege system. )


stelmo (approved):
More power to the people

PR #645, opened 2022-08-01
unify the reference section with the rest of the docs

exaexa:


docs/src/functions.md
```diff
@@ -1,4 +1,4 @@
-# Types and functions
+# API reference
```
stelmo, line 2: A detailed listing of the functions can be seen here (link)

stelmo: Done below!

stelmo (approved):

docs/src/functions.md
exaexa: yep that's in another file

PR #648, opened 2022-08-02
genetic manipulation wrappers

### LINKED ISSUE #552: Model type overlay that describes the genetic manipulation
Copied from #520:

Folding the logic into data is right. Actually I was contemplating on something similar already with #510.

Technically we can do an "overlay" type that just adds the functionality to the inner type, just as CoreModelCoupled adds the coupling or Serialized adds the disk cache.

Random idea:

```julia
struct WithGeneActivation{M :> MetabolicModel}
  m::M
  modified_genes::Dict{String,Float64}
end
```

_Originally posted by @exaexa in https://github.com/LCSB-BioCore/COBREXA.jl/issues/520#issuecomment-974107433_

exaexa:
## Description

This should subsume various functionality that was drafted previously but not merged (EFlux) and also a generic way to add knockouts to models.

Closes #552 . 
(eventually)

In the process, I'll break `knockout` to clean up the naming, so this really needs to go to 2.0. Which is otherwise not a big deal because the naming of stuff is going to get revamped there anyway.


exaexa:
deprecated by #777

PR #649, opened 2022-08-05
Remove the banner

stelmo:
The banner is totally unnecessary + a cobratoolbox left over, 🪓 it for v2.0.


exaexa (approved):
After some discussion and a bit of tears about the banner having :sob: to go I'm OK with this.

Still, could we actually move the banner and version information into the module docs?

Like, if people type `?COBREXA`, this (and the version information, etc.) could show up instead of the really completely useless README print-out.

exaexa (commented):
there was a `COBREXA_VERSION` const defined in the banner, which was not internal and people (incl me) might have depended on it. I think it would be useful to have that preserved. Not sure if it can be reused within the module docstring though.

laurentheirendt:
I still don't think this is nice ...

exaexa (approved):

PR #654, opened 2022-08-27
Add some constructors

stelmo:
Useful for model building. Basically moved the existing internal constructors outside so that they can be documented and appear when you look for them at the repl.

### DIFF
--- a/src/base/types/Gene.jl
+++ b/src/base/types/Gene.jl
@@ -12,12 +12,7 @@ mutable struct Gene
 end
 
 """
-Gene(
-    id=""; 
-    name = nothing, 
-    notes = Notes(), 
-    annotations = Annotations(),
-)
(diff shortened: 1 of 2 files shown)


exaexa (commented):
This should be rebased when the DSE PR is merged. (#661)

Also, shouldn't we use `Base.@kwdef` or some of the pals as we do with SBML? It makes this kind of constructors automagically.

Ref: https://github.com/LCSB-BioCore/SBML.jl/blob/master/src/structs.jl#L413

stelmo:
Oh yeah, `Base.kwdef` will be good to add. Okay, will do once rebased

stelmo:
Ah actually, I want the docstrings, not sure if kwdef adds docstrings by default?

exaexa (approved):
cool

PR #656, opened 2022-08-29
hotfix: bump CompatHelper version to work with 1.8

exaexa:
Same issue and solution as with https://github.com/LCSB-BioCore/GigaSOM.jl/pull/206

For CI reasons I think merging right into master will be much better here (feel free to reparent if not).

### DIFF
--- a/.github/workflows/CompatHelper.yml
+++ b/.github/workflows/CompatHelper.yml
@@ -12,7 +12,7 @@ jobs:
           import Pkg
           name = "CompatHelper"
           uuid = "aa819f21-2bde-4658-8897-bab36330d9b7"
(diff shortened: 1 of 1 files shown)


laurentheirendt (approved):

laurentheirendt:
shouldn't this be going into `develop`?

exaexa:
> shouldn't this be going into `develop`?

no, let's do `master`. If we merge into develop we basically postpone fix until a new release is out (which is undesirable now because we have breaking changes pending), or until we do a slighly non-systematic develop->master merge.

exaexa:
Develop counterpart: #658

PR #658, opened 2022-08-30
Update compathelper CI script

exaexa:
a more complete version of hotfix #656 (refreshes the whole CompatHelper script)

### DIFF
--- a/.github/workflows/CompatHelper.yml
+++ b/.github/workflows/CompatHelper.yml
@@ -12,7 +12,7 @@ jobs:
           import Pkg
           name = "CompatHelper"
           uuid = "aa819f21-2bde-4658-8897-bab36330d9b7"
-          version = "2"
+          version = "3"
           Pkg.add(; name, uuid, version)
         shell: julia --color=yes {0}
       - name: "Run CompatHelper"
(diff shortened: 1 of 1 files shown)


.github/workflows/CompatHelper.yml
```diff
@@ -3,16 +3,35 @@ on:
...
     - cron: 0 0 * * *
   workflow_dispatch:
+permissions:
+  contents: write
+  pull-requests: write
 jobs:
   CompatHelper:
     runs-on: ubuntu-latest
     steps:
+      - name: Check if Julia is already available in the PATH
```
laurentheirendt, line 13: Why are these here?

exaexa: it is a clone of the officially recommended CH script:
https://github.com/JuliaRegistries/CompatHelper.jl/blob/master/.github/workflows/CompatHelper.yml

laurentheirendt: Are they from a previous hot fix?

exaexa: (apparently github may provide some julia in base images)

laurentheirendt: okay, but then this will go only later to `master` then ....

laurentheirendt (approved):

PR #660, opened 2022-08-30
Trial of using Term.jl as our pretty printing interface

stelmo:
We should consider using [Term.jl](https://fedeclaudi.github.io/Term.jl/stable/) to pretty print stuff instead of our manual way. I think it looks pretty cool because it gives you a lot of information about the struct for free. This is what standardmodel would look like using term. 
image

I did notice that there was a significant compile time when displaying this... So that is something to consider...


exaexa:
I guess we'll need to finetune most of the output to be useful, e.g. in the picture you shared:
- +60% space is wasted on types and incomplete information
- the basic model info (sizes etc) we print out now is missing

stelmo:
🪓 this, custom pretty printing for now!

exaexa:
:sob: tho.

PR #661, opened 2022-08-30
Finally add DocStringExtensions 🥳 

stelmo:
This should reduce docstring out-of-sync issues!

### DIFF
--- a/src/base/logging/log.jl
+++ b/src/base/logging/log.jl
@@ -1,6 +1,6 @@
 
 """
-    macro _make_logging_group(sym::Symbol, doc::String)
+$(TYPEDSIGNATURES)
 
 This creates a group of functions that allow masking out topic-related logging
 actions. A call that goes as follows:
--- a/src/base/macros/model_wrapper.jl
+++ b/src/base/macros/model_wrapper.jl
@@ -38,7 +38,7 @@ function _inherit_model_methods_impl(
 end
 
 """
-    @_inherit_model_methods
+$(TYPEDSIGNATURES)
 
 Generates trivial accessor functions listed in `fns` for a model that is
 wrapped in type `mtype` as field `member`.
@@ -55,7 +55,7 @@ macro _inherit_model_methods(mtype::Symbol, arglist, member::Symbol, fwdlist, fn
 end
 
 """
-    @_inherit_model_methods_fn
+$(TYPEDSIGNATURES)
 
 A more generic version of [`@_inherit_model_methods`](@ref) that accesses the
 "inner" model using an accessor function name.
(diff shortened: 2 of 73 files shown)


laurentheirendt:
this is cool 👍🏼

src/base/logging/log.jl
```diff
@@ -1,6 +1,6 @@
 
 """
-    macro _make_logging_group(sym::Symbol, doc::String)
+$(TYPEDSIGNATURES)
```
exaexa, line 3: Q: is there any way to use TYPEDSIGNATURES in the macro that actually generates the functions (below in this file)?

src/base/types/abstract/Maybe.jl
```diff
@@ -7,7 +7,7 @@ A nice name for "nullable" type.
 const Maybe{T} = Union{Nothing,T}
```
exaexa, line 7: Q: why no DSE here?

src/base/macros/model_wrapper.jl
```diff
@@ -1,6 +1,6 @@
 
 """
-    _inherit_model_methods_impl(mtype::Symbol, arglist, access, fwdlist, fns...)
+$(TYPEDSIGNATURES)
 
 A helper backend for [`@_inherit_model_methods`](@ref) and
 [`@_inherit_model_methods_fn`](@ref).
```
exaexa, line 6: Q: the macros lower in this file can't be documented with DSE?

(Also, would it be possible to use DSE again for the generated docstring below?)

exaexa (commented):
A few questions but otherwise looks okay

extra questions:
- what about the extra operators in `src/reconstruction/Reactions.jl`? (could we have the typed signature and print out the intended inline usage in an example?)
- what about code generated by macros, is it possible to smuggle TYPEDSIGNATURES in there?

src/base/types/abstract/Maybe.jl
stelmo: I did not DSE anything constants, not sure they have something for it?

stelmo:
> Ok this is a good job. I've got a few questions but otherwise looks okay.
> 
> * what about the extra operators in `src/reconstruction/Reactions.jl`? (could we have the typed signature and print out the intended inline usage in an example?)

They looked weird so I left them, but tbh, there is no good reason for leaving them. Adding them! 

> * what about code generated by macros (e.g. in `src/base/macros/model_wrapper.jl`), is it possible to smuggle `TYPEDSIGNATURES` in there?

Inserting docstrings inside a macro that generates documentation seems to be an issue..

> * is there no way to print out the `const` typedefs (e.g. in `src/base/types/abstract/MetabolicModel.jl`)

I did not find anything for constants... Seems they will just be manual

src/base/types/abstract/Maybe.jl
exaexa: ah this is the case of "constants docs", nvm.

exaexa (approved):
All answered I guess. Re the operators, since they are user-facing and the involved types are completely internal I guess we can do without the precise doc.

PR #667, opened 2022-09-03
Add SBO terms

stelmo:
SBO is a systematic naming convention. We should use it to look for reactions instead of guessing based on naming. Right now just add the ontology and a few `is_xxx` functions, but for 2.0 I propose to only using sbo terms in functions like `find_exchange_reactions`. This removes the footgun of assuming `find_xxx` is robust (it isn't because it depends on `looks_like`)


src/base/sbo.jl
```diff
@@ -0,0 +1,80 @@
+#=
+This file contains a list of SBO terms recognized by COBREXA. For the full
+ontology, see https://github.com/EBI-BioModels/SBO/blob/master/SBO_OBO.obo. 
+
+If an SBO term appears here it *may* be used in a function; if an SBO term does
+not appear here, then it is *not* used in a COBREXA function.
+=#
+
+const FLUX_BALANCE_FRAMEWORK = "SBO:0000624"
```
exaexa, line 9: What about making the stuff slightly more discoverable right away?

```suggestion
module SBOTerms
 const FLUX_BALANCE_FRAMEWORK = "SBO:0000624"
...
```
or perhaps
```suggestion
const SBOTerms = Dict(
  "flux balance framework" => "SBO:0000624",
  ...
```
I'm a slighly bigger fan of the 2nd approach because it also allows us to translate SBO terms back to reasonable strings. Also we should not invent random `SCREAMING_SNAKE` constants and follow the precise naming from SBO ontology, which contains spaces so it's not very good for actual identifiers.

exaexa (approved):
lgtm

Issue #657, opened 2022-08-29
Sampling size parameters are not very intuitive


stelmo:
Why does the sampling procedure generate as many samples as one inputs warmup points? Shouldn't it return as many samples as requested in `sample_iters`? See below for a MWE:
```julia
using COBREXA, Gurobi

model = load_model(StandardModel, joinpath("data", "models", "e_coli_core.json"))

warmup_points = warmup_from_variability(
    model, 
    Gurobi.Optimizer,
    10,
    modifications = [silence]
)

samples = affine_hit_and_run(
    model,
    warmup_points,
    sample_iters = 1:1:5
)
```
I expect 5 samples here, but I get 5 * 10 = 50. Is this expected behaviour?

exaexa:
solved in cobrexa 2

PR #668, opened 2022-09-03
Change how samples are collected

stelmo:
Currently each sample collection iteration collects as many samples as there are warmup points. I think it is more intuitive to collect samples only as specified by `sample_iters` and let `chains` do the work of generating batches. This simplifies the current the implementation. This is possibly breaking as a smaller output dimension is returned (not sure if this counts)...

### DIFF
--- a/src/analysis/sampling/affine_hit_and_run.jl
+++ b/src/analysis/sampling/affine_hit_and_run.jl
@@ -1,25 +1,27 @@
 """
 $(TYPEDSIGNATURES)
 
-Run a hit-and-run style sampling that starts from `warmup_points` and uses
-their affine combinations for generating the run directions to sample the space
-delimited by `lbs` and `ubs`.  The reaction rate vectors in `warmup_points`
-should be organized in columns, i.e. `warmup_points[:,1]` is the first set of
-reaction rates.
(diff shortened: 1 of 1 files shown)


stelmo:
If merged, fixes and closes #657

exaexa (commented):
minor issues with this:
- it is not really `affine` anymore because the affine combinations are gone.
- for backwards compatibility the defaults should give the same output size as the previous version's default (you can schedule breaking it for 2.0 later)

stelmo:
I put the breaking tag on it because this is technically slightly breaking

exaexa (changes requested):
OK so

can we rename the "affine" to something else? there's no affine combinations in this one so the name becomes really misleading. The original functionality I will retain because it's a base for the CUDA version (which I should finally release, right?)

stelmo:
Hmm we need a name for it... While `hit_and_run` is not technically correct (because our directions are not truly random, but based on sampled points), I think it is a good enough name and relatively accepted in the field...

stelmo:
Can I close this? all sampling goes to your sampling repo?

exaexa:
it didn't solve the immediate problem (the whole relation to CuFluxSamplers should be cleared out) but yeah we can....

PR #672, opened 2022-09-22
Re-organize everything into modules to make discoverability easier

stelmo:
I also got rid of the autoloading "feature", it gets in the way more than it helps... 

Possible modules include:
1. Common
2. IO 
3. Analysis
4. Reconstruction

Open questions:
1. How many levels deep do we want to go? Like `COBREXA.Common.Types` or just `COBREXA.Common`
2.  tbd


stelmo:
@exaexa I am having trouble with the macros, they don't exist when they need to, do you know how to make the current state of the PR compile? Also, how do I use `FromFile` correctly? Or is the multiple importing thing okay?

stelmo:
Currently the problem is with the wrappers. They have helper functions (currently located in `types/wrappers/SMomentModel.jl` or `types/wrappers/GeckoModel.jl`) which require the accessors to be used, but they are not accessors... :/

stelmo:
Instead use #674

exaexa:
:headstone:  tho.

PR #676, opened 2022-09-27
Make the SBML annotations work just like with JSON models

exaexa:
This imports CV terms from SBML1.3 (and saves them back). Import and representation methodology is moreless the same as with cobra.:snake: . Because finally fixing this I bump 1.4.1.

There is a bunch of minor, partially surprising fixes that were packed along the way. Most notably there's no realistically good way to export notes to SBML 

Cc @HettieC -- this should finally fix the whole annotation problem.


stelmo (approved):
Nice

PR #680, opened 2022-10-13
switch tests julia from 1.7 to 1.8

exaexa:
1.7 is old "stable" now; this should migrate the tests to LTS (1.6) + newest Julia (1.8)

I expect trouble though :sweat_smile:


stelmo (approved):
looks good

PR #687, opened 2022-10-25
Rename `MetabolicModel` to `AbstractMetabolicModel`

stelmo:
Following Julia style by calling abstract types with the prefix `Abstract`


exaexa (commented):
what about if I do also the other renames right here?

exaexa (approved):
cool.

PR #689, opened 2022-10-27
Update gene associations

stelmo:
Slightly improve gene associations. Will break almost everything. This also completely wrecks how we build gecko models. But this will be fixed within the next few weeks


src/types.jl
```diff
@@ -36,35 +36,35 @@ Overload these if you want COBREXA to work with your own module types.
...
-using .Internal
+    using .Internal
 
-@inc_dir types accessors
-@export_locals
+    @inc_dir types accessors
+    @export_locals
 end
 
 # the modules depend on each other so we have to inject the stuff like this
```
stelmo, line 70: I added these spaces to make it more readable.. I didn't know where the module begins or ends

src/types/Gene.jl
```diff
@@ -4,17 +4,10 @@ $(TYPEDEF)
 # Fields
 $(TYPEDFIELDS)
 """
-mutable struct Gene
+Base.@kwdef mutable struct Gene
```
stelmo, line 7: Put this in a lot of places, slightly unrelated to the PR, but nice to use!

src/types/misc/gene_associations.jl
```diff
@@ -2,10 +2,10 @@
 """
 $(TYPEDSIGNATURES)
 
-Parse `SBML.GeneProductAssociation` structure to the simpler GeneAssociation.
+Parse `SBML.GeneProductAssociation` structure to the simpler [`GeneAssociations`](@ref).
```
stelmo, line 5: Changed it to plural because it is a vector now

src/types/models/ObjectModel.jl
```diff
@@ -35,18 +32,12 @@ keys(model.reactions)
...
-    ObjectModel(
-        id = "";
-        reactions = OrderedDict{String,Reaction}(),
-        metabolites = OrderedDict{String,Metabolite}(),
-        genes = OrderedDict{String,Gene}(),
-    ) = new(id, reactions, metabolites, genes)
+    reactions::OrderedDict{String,Reaction} = OrderedDict{String,Reaction}()
+    metabolites::OrderedDict{String,Metabolite} = OrderedDict{String,Metabolite}()
+    genes::OrderedDict{String,Gene} = OrderedDict{String,Gene}()
+    objective::Maybe{Tuple{String, Float64}} = nothing
```
exaexa, line 40: what about multi-reaction objectives?

src/types/models/ObjectModel.jl
```diff
@@ -35,18 +32,12 @@ keys(model.reactions)
 # Fields
 $(TYPEDFIELDS)
 """
-mutable struct ObjectModel <: AbstractMetabolicModel
+Bas.@kwdef mutable struct ObjectModel <: AbstractMetabolicModel
```
exaexa, line 35: kwdef :partying_face:

src/types/Gene.jl
exaexa: yeah certainly helps but we should have a PR that does it properly and globally. E.g. here you can move the defaults from the constructor to the struct definition.

exaexa: ( oh okay you did already, good. :D )

src/types.jl
exaexa: Is there perhaps a JuliaFormatter setting for that? If this is not reproducible we'll get the PRs pwned by formatchecker forever.

exaexa (changes requested):
can we have the tests fixed? Also please keep this formatted, I'll do the module formatting change in another PR

src/types.jl
stelmo: will use the new formatter settings

src/types/models/ObjectModel.jl
stelmo: changed to `Maybe{Vector{Tuple{String, Float64}}}`

exaexa: So why not `Dict{String,Float64}` ?

stelmo: yeah, that is better

src/types/misc/gene_associations.jl
```diff
@@ -2,10 +2,10 @@
...
 The input must be (implicitly) in a positive DNF.
 """
-function parse_grr(gpa::SBML.GeneProductAssociation)::GeneAssociation
+function parse_grr(gpa::SBML.GeneProductAssociation)
```
exaexa, line 8: pls do not erase return types, they serve as a sanity check before stuff gets totally broken later in the pipeline.

Also the original GeneAssociation might be better renamed to GeneAssocDNF or something like that.

src/types/models/MatrixModel.jl
```diff
@@ -21,7 +21,7 @@ mutable struct MatrixModel <: AbstractMetabolicModel
     xu::Vector{Float64}
     rxns::Vector{String}
     mets::Vector{String}
-    grrs::Vector{Maybe{GeneAssociation}}
+    grrs::Vector{Maybe{Vector{Vector{String}}}}
```
exaexa, line 24: Why doesn't this use the isozmes? (in the other case use GeneAssocDNF or a similar type name, this is bad)

src/types/models/ObjectModel.jl
```diff
@@ -35,18 +32,12 @@ keys(model.reactions)
...
-        id = "";
-        reactions = OrderedDict{String,Reaction}(),
-        metabolites = OrderedDict{String,Metabolite}(),
-        genes = OrderedDict{String,Gene}(),
-    ) = new(id, reactions, metabolites, genes)
+    reactions::OrderedDict{String,Reaction} = OrderedDict{String,Reaction}()
+    metabolites::OrderedDict{String,Metabolite} = OrderedDict{String,Metabolite}()
+    genes::OrderedDict{String,Gene} = OrderedDict{String,Gene}()
+    objective::Dict{String,Float64} = Dict{String,Float64}()
 end
```
exaexa, line 41: btw you may want to check SBML source -- iirc you can actually add docstrings to fields which get rendered as documentation in $(TYPEDFIELDS)

src/reconstruction/MatrixModel.jl
```diff
@@ -21,11 +21,10 @@ function add_reactions!(model::MatrixModel, rxns::Vector{Reaction})
         push!(model.rxns, rxn.id)
         lbs[j] = rxn.lower_bound
         ubs[j] = rxn.upper_bound
-        cs[j] = rxn.objective_coefficient
     end
     Sadd = sparse(I, J, V, n_metabolites(model), length(rxns))
     model.S = [model.S Sadd]
-    model.c = dropzeros([model.c; cs])
+    model.c = spzeros(size(model.S, 2)) # does not add an objective
```
exaexa, line 27: so this suddenly drops the original objective? that sounds a bit harsh

exaexa (changes requested):
minor cosmetic stuff and 1 surprise but looks pretty much good to go

src/types/models/ObjectModel.jl
exaexa: see e.g. here:
https://github.com/LCSB-BioCore/PikaParser.jl/blob/master/src/structs.jl#L311
and rendering here:
https://lcsb-biocore.github.io/PikaParser.jl/stable/reference/#PikaParser.ParserState

src/types/models/MatrixModel.jl
stelmo: So isozymes is an ObjectModel only type of struct. Will add something like GeneAssocDNF  to represent those vector vector strings

src/types/models/ObjectModel.jl
stelmo: ahhh, okay, will make OM prettier

src/reconstruction/MatrixModel.jl
stelmo: fixed

exaexa (commented):
lgtm now

exaexa (approved):

PR #695, opened 2022-11-03
Add accessor to get model IDs

stelmo:


stelmo:
This should be moved to FBCModelTests instead, generic ID accessors =  feature creep

PR #699, opened 2022-11-10
Turn MMDF analysis into a wrapper

### LINKED ISSUE #614: `max_min_driving_force` and `moment` should be a `ModelWrapper`
## Target functionality

To make MMDF easily cooperate with the rest of the ecosystem, it should be expressed as a separate model wrapper.

## Desired output

For example, that I can combine MMDF with sMOMENT by just adding the 2 wraps.

## Suggestions for implementation

Since there's only a single solve needed, this should be easily possible to do just by following the sMOMENT or GECKO wrappers that are already present.

stelmo:
Wrappers fit in more with the COBREXA ecosystem. Let's convert all the MMDF analyses into a `ThermodynamicModel` that is a wrapper. 

Closes #614 when complete


stelmo:
#700 needs to be done before this one

stelmo:
NB: merge only after #742

src/types/accessors/AbstractMetabolicModel.jl
```diff
@@ -141,6 +141,25 @@ groups collect these sets of enzymes for convenient analysis.
...
+variables. This semantic grouping uses the log (base e) metabolite concentration
+to make thermodynamic calculations easier.
+"""
+)
+
+@make_variable_semantics(
+    :gibbs_free_energy_reaction,
+    "Gibbs free energy of reaction",
+    """
+Thermodynamic models need to ensure that the ΔG of each reaction is negative
```
exaexa, line 158: "Some thermodynamic models..."

src/types/accessors/AbstractMetabolicModel.jl
```diff
@@ -141,6 +141,25 @@ groups collect these sets of enzymes for convenient analysis.
...
+    "metabolite log concentration",
+    """
+Certain model types use metabolite concentrations instead of reaction fluxes are
+variables. This semantic grouping uses the log (base e) metabolite concentration
+to make thermodynamic calculations easier.
+"""
+)
+
+@make_variable_semantics(
+    :gibbs_free_energy_reaction,
```
exaexa, line 155: this is a good dodge of the "energys". We should really have `gibbs_free_energy` and `gibbs_free_energies`. I'll make a macro update that fixes this.

src/types/accessors/AbstractMetabolicModel.jl
```diff
@@ -141,6 +141,25 @@ groups collect these sets of enzymes for convenient analysis.
 """
 )
 
+@make_variable_semantics(
+    :metabolite_log_concentration,
```
exaexa, line 145: can we generalize to log-concentration to anything, or should this stay metabolite specific? (just golfing the name)

src/types/wrappers/MaxMinDrivingForceModel.jl
```diff
@@ -0,0 +1,307 @@
...
+
+# Fields
+$(TYPEDFIELDS)
+"""
+Base.@kwdef mutable struct MaxMinDrivingForceModel <: AbstractModelWrapper
+    "A dictionary mapping ΔrG⁰ to reactions."
+    reaction_standard_gibbs_free_energies::Dict{String,Float64} = Dict{String,Float64}()
+
+    "A cycle-free reference flux solution that is used to set the directions of the reactions."
+    flux_solution::Dict{String,Float64} = Dict{String,Float64}()
```
exaexa, line 39: this should document how to get the reference solution. Is it variable values or reaction flux values or something else? Can this work on more complicated reactions? (If not we're all ok, just make a note that "this can't yet work on multivariable/split reactions")

src/types/wrappers/MaxMinDrivingForceModel.jl
```diff
@@ -0,0 +1,307 @@
...
+"""
+Accessors.variables(model::MaxMinDrivingForceModel) =
+    ["mmdf"; "log " .* metabolites(model); "ΔG " .* reactions(model)]
+
+"""
+$(TYPEDSIGNATURES)
+
+Helper function that returns the unmangled variable IDs.
+"""
+get_unmangled_variables(model::MaxMinDrivingForceModel) =
```
exaexa, line 96: `mmdf_original_variables`

src/types/wrappers/MaxMinDrivingForceModel.jl
```diff
@@ -0,0 +1,307 @@
...
+    ["mmdf"; metabolites(model); reactions(model)]
```
exaexa, line 97: `reactions` or `variables` ?

src/types/wrappers/MaxMinDrivingForceModel.jl
exaexa: (btw might be better shoveled something into `Internal` or so, so that users can't access it)

src/types/wrappers/MaxMinDrivingForceModel.jl
```diff
@@ -0,0 +1,307 @@
...
+end
+
+"""
+$(TYPEDSIGNATURES)
+
+A helper function that returns the reaction ids that are active. Active reaction
+have thermodynamic data AND a flux bigger than `small_flux_tol` AND are not
+ignored.
+"""
+_get_active_reaction_ids(model::MaxMinDrivingForceModel) = filter(
```
exaexa, line 251: underscores are obsolete, `get_` prefix is imperative...

...move this to Internal, like `mmdf_active_reaction_ids` or so

exaexa (commented):
Pretty much OK. I'll need to fix the semantics-generating macro to allow sensible plurals for `energys` etc., then I'll merge it. There are some notes for code moving/reorganization, that should be easy. If

src/types/accessors/AbstractMetabolicModel.jl
stelmo: actually in this case you don't need to, because the value calculated is literally the Gibbs free energy of reaction

src/types/accessors/AbstractMetabolicModel.jl
stelmo: hmm probably only for metabolites does it make sense. Other concentration like quantities will likely not get logged since they won't feature into thermo calculations e.g. enzyme concentrations

src/types/wrappers/MaxMinDrivingForceModel.jl
stelmo: reactions

src/types/accessors/AbstractMetabolicModel.jl
stelmo: yeah it would be nice to group all the semantics somewhere for easy reference

exaexa (approved):
lovely. Let's not care about the gibbs issue for now.

PR #700, opened 2022-11-11
Change the interface to the optimizer

stelmo:
Change the optimization interface to make the type of problems handled by COBREXA more general.

### DIFF
--- a/src/analysis/flux_balance_analysis.jl
+++ b/src/analysis/flux_balance_analysis.jl
@@ -1,40 +1,6 @@
 """
 $(TYPEDSIGNATURES)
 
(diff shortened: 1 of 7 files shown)


exaexa (commented):
Some of this has been already merged (renaming of variables).

Leftovers to take out:
- quadratic objectives (easy and nonbreaking but let's do it now so we can fix other stuff)
- `Result`. The parameter type should not be a tag (these are not extensible), let's go with plain type at first, but inventing some kind of tagging would be great (so that people e.g. don't confuse a sample with optimum or so).

stelmo:
This is addressed in #736

PR #701, opened 2022-11-18
use a softer failure when unit annotations are missing in SBML

exaexa:


stelmo (approved):
Looks good to me! Weird that = worked instead of in!

PR #705, opened 2022-11-25
upgrade julia version on ci and set mac path

laurentheirendt:
## Description

This upgrades the julia version on the CI and updates the binary path on the macOS node.


exaexa (approved):
perfect thanks

PR #706, opened 2022-12-06
freeze SBML dictionary ordering

exaexa:
Hashtable ordering is more random than was expected; and the unpredictability caused bad/random results when processing large models in a distributed way.

This forcibly freezes the order for SBMLModel so that the ordering issue is avoided completely and forever.


exaexa:
(I had to intern some code from SBML.jl, but hopefully for good.)

src/base/types/SBMLModel.jl
```diff
@@ -56,21 +115,56 @@ Get the lower and upper flux bounds of model [`SBMLModel`](@ref). Throws `Domain
...
-
-    unit = lbu[1][2]
-    getvalue = (val, _)::Tuple -> val
-    getunit = (_, unit)::Tuple -> unit
+    # There are multiple ways in SBML to specify a lower/upper bound. There are
+    # the "global" model bounds that we completely ignore now because no one
+    # uses them. In reaction, you can specify the bounds using "LOWER_BOUND"
+    # and "UPPER_BOUND" parameters, but also there may be a FBC plugged-in
+    # parameter name that refers to the parameters.  We extract these, using
+    # the units from the parameters. For unbounded reactions we use -Inf or Inf
```
stelmo, line 123: should we use the reaction bound constants here?

stelmo (approved):
Looks good to me, but we should talk about Dictionaries.jl :)

PR #710, opened 2022-12-15
huge rename of variable-ish things

exaexa:
Bonuses:
- reaction_names and similar now looks better
- somehow the tests run OK for me

Possibly unsolved:
- what do we call the metabolites (in general)?


stelmo (approved):
severe emotional disturbance ftw

PR #712, opened 2022-12-21
Fix a small bug + add 2 convenience functions

stelmo:
These 2 convenience functions will get changed with #700, but for now it is useful :)

### DIFF
--- a/src/utils/BalancedGrowthCommunityModel.jl
+++ b/src/utils/BalancedGrowthCommunityModel.jl
@@ -1,0 +1,22 @@
+"""
+$(TYPEDSIGNATURES)
+
+Extract the solution of a specific `community_member` from `opt_model`, which is
+a solved optimization model built from the `community_model`. Removes the
+`community_member` prefix in the string ids of the returned dictionary.
+"""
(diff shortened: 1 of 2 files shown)


src/utils/BalancedGrowthCommunityModel.jl
```diff
@@ -0,0 +1,22 @@
...
+) : nothing
+
+
+"""
+$(TYPEDSIGNATURES)
+
+Extract the environmental exchanges from `opt_model`, which is a solved
+optimization model built from the `community_model`.
+"""
+get_environmental_exchanges(opt_model, community_model::BalancedGrowthCommunityModel) = is_solved(opt_model) ? Dict(
```
exaexa, line 19: consider keeping the arguments in a consistent order, preferably sorted by "logical size" ( model ≥ optmodel ≥ piece of model ).

Other than that I'm okay with this.

exaexa (approved):
1 suggestion worthy considering, other than that OK

src/utils/BalancedGrowthCommunityModel.jl
stelmo: cool, will change!

PR #717, opened 2023-01-07
Implement a variant of Constrained Allocation Flux Balance Analysis

stelmo:
CAFB incorporates the so-called bacterial growth law into enzyme constrained FBA. By adding a wrapper on top of Gecko and SMoment we can pretty easily incorporate the gist of CAFB.


src/types/wrappers/xConstrainedAllocationModel.jl
```diff
@@ -0,0 +1,124 @@
...
+"""
+$(TYPEDSIGNATURES)
+
+Helper function that adds the ribosome trade-off to a [`SMomentModel`](@ref).
+"""
+function _cam_smoment(model::ConstrainedAllocationModel)
+    _idx = first(indexin([model.growth_rate_id], reactions(model.inner)))
+    isnothing(_idx) && throw(error("Growth rate reaction ID not found."))
+    # need to adjust idx using smoment_column_reactions due to split reactions (assume biomass reaction is not reversible)
+    cols = smoment_column_reactions(model.inner)
```
exaexa, line 117: once we've got variables, can't this just be gathered from stoichiometry?
...also I should rename to _variables :]

src/types/wrappers/xConstrainedAllocationModel.jl
```diff
@@ -0,0 +1,124 @@
...
+    idx = first(findall(!iszero, cols[_idx, :]))
```
exaexa, line 118: how's this idx different from _idx ? might be better name something like `growth_reaction_idx` or `growth_metabolite_idx` or something, not sure what it is now.

src/types/wrappers/xConstrainedAllocationModel.jl
```diff
@@ -0,0 +1,124 @@
...
+
+    caps = [col.capacity_required for col in model.inner.columns]
+    caps[idx] = model.weight
+
+    vcat(coupling(model.inner.inner) * cols, caps')
```
exaexa, line 123: why model.inner.inner? the coupling of gecko/smoment is omitted?

src/types/wrappers/xConstrainedAllocationModel.jl
```diff
@@ -0,0 +1,124 @@
...
+    inner::Union{SMomentModel,GeckoModel}
+
+    ConstrainedAllocationModel(weight, growth_rate_id, total_enzyme_bound_id, inner) = begin
+        inner isa GeckoModel ||
+        inner isa SMomentModel ||
+        throw(error("Model type unsupported by ConstrainedAllocationModel."))
+        idx = first(indexin([growth_rate_id], reactions(inner)))
+        lbs, ubs = bounds(inner)
+        lbs[idx] * ubs[idx] < 0 &&
+        throw(error("Growth rate reaction must be unidirectional."))
```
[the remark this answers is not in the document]
stelmo: Why would this be better?

src/types/wrappers/xConstrainedAllocationModel.jl
```diff
@@ -0,0 +1,124 @@
...
+[`SMomentModel`](@ref) the field `total_enzyme_bound_id` does not need to be
+specified.
+
+See Mori, Matteo, et al. "Constrained allocation flux balance analysis." PLoS
+computational biology 12.6 (2016) for more details.
+
+# Fields
+$(TYPEDFIELDS)
+"""
+struct ConstrainedAllocationModel <: AbstractModelWrapper
```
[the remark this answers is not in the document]
stelmo: ah yes! I forgot about this

src/types/wrappers/xConstrainedAllocationModel.jl
```diff
@@ -0,0 +1,124 @@
...
+$(TYPEDSIGNATURES)
+
+Return the coupling of [`ConstrainedAllocationModel`](@ref). In addition to the
+coupling constraints of the underlying model, this adds the ribosome trade-off
+to the total proteome capacity bound.
+"""
+function Accessors.coupling(model::ConstrainedAllocationModel)
+    model.inner isa GeckoModel && return _cam_gecko(model)
+    model.inner isa SMomentModel && return _cam_smoment(model)
+end
```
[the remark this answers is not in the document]
stelmo: yeah, we should go through cobrexa at some stage and add better error messages

src/types/wrappers/xConstrainedAllocationModel.jl
stelmo: my bad, it should be model.inner

stelmo: actually, this just reconstructs the coupling from the inner.inner model

src/types/wrappers/xConstrainedAllocationModel.jl
```diff
@@ -0,0 +1,124 @@
...
+    ]
+end
+
+"""
+$(TYPEDSIGNATURES)
+
+Helper function that adds the ribosome trade-off to a [`SMomentModel`](@ref).
+"""
+function _cam_smoment(model::ConstrainedAllocationModel)
+    _idx = first(indexin([model.growth_rate_id], reactions(model.inner)))
```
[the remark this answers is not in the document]
stelmo: variables, but is that already in `next`?

src/types/wrappers/xConstrainedAllocationModel.jl
stelmo: I don't think so, you need to know which column to adjust in the last row of the coupling matrix (that is where the total capacity constraint is). It's position depends on the expanded variables (which include the split forward and reverse reactions)

src/types/wrappers/xConstrainedAllocationModel.jl
stelmo: renamed it

src/types/wrappers/xConstrainedAllocationModel.jl
```diff
@@ -0,0 +1,128 @@
...
+"""
+$(TYPEDSIGNATURES)
+
+Return the coupling of [`ConstrainedAllocationModel`](@ref). In addition to the
+coupling constraints of the underlying model, this adds the ribosome trade-off
+to the total proteome capacity bound.
+"""
+function Accessors.coupling(model::ConstrainedAllocationModel)
+    model.inner isa GeckoModel && return _cam_gecko(model)
+    model.inner isa SMomentModel && return _cam_smoment(model)
```
exaexa, line 79: underscores no longer needed

stelmo:
New idea: convert this into a constructor that merely adds a ribosome protein to a model so that everything "just works" with the current gecko and smoment

stelmo:
Note: merge this after #724

src/reconstruction/ObjectModel.jl
```diff
@@ -234,3 +241,180 @@ function add_biomass_metabolite(
...
+    nothing
+end
+
+"""
+$(TYPEDSIGNATURES)
+
+Variant of [`add_pseudoribosome!`](@ref) that does not modify the original
+model, but makes a shallow copy with the modification included.
+"""
+function add_pseudoribosome(
```
exaexa, line 311: virtual (too many actual real stuff in biology is called "pseudosomething")

src/reconstruction/modifications/enzymes.jl
```diff
@@ -17,3 +17,20 @@ giving a [`EnzymeConstrainedModel`](@ref). The arguments are forwarded to
...
+with_pseudoribosome(args...; kwargs...) =
+    model -> add_pseudoribosome(model, args...; kwargs...)
+
+"""
+$(TYPEDSIGNATURES)
+
+Specifies a model variant that overwrites the current isozymes associated with the
+model through calling [`add_isozymes`](@ref).
+"""
+with_isozymes(args...) = model -> add_isozymes(model, args...)
```
exaexa, line 36: this only works on the ObjectModel now but whatevs; eventually we're going to generalize that to all models (with the wrappers) and this code stays intact.

As the only change pls forward kwargs too for future compat (even though there are none now). I'd even avoid having any positional args in the first place, luckily here it is kinda intuitive -- here it says you're adding isozymes so the only arg is likely gonna be the isozymes. :D

exaexa (changes requested):
Small changes but otherwise OK for `next`

src/reconstruction/ObjectModel.jl
stelmo: so technically virtual is more accurate than pseudo! changing

PR #721, opened 2023-01-09
Rename reaction_gene_association to reaction_gene_associations

stelmo:
This is more consistent since we use `xxx_notes` and `xxx_annotations` and the rgas are vectors


exaexa (approved):
OK

PR #722, opened 2023-01-09
Replace uses of `:reverse` with `:backward` to be consistent

stelmo:


exaexa (approved):
this is very backward

PR #725, opened 2023-01-11
Remove crowding (obsolete via SMOMENT)

stelmo:
Crowding is obsolete since the inclusion of smoment (just add random or equal weights to all enzymes for the same effect).

Moreover, it should not be a modification.


docs/src/examples/10_ADD_ANOTHER_EXAMPLE.jl
```diff
@@ -0,0 +1,13 @@
...
+
+!isfile("e_coli_core.xml") &&
+    download("http://bigg.ucsd.edu/static/models/e_coli_core.xml", "e_coli_core.xml")
+
+using COBREXA, Tulip
+
+model = load_model("e_coli_core.xml")
+
+import Random
+Random.seed!(1) # for repeatability of random numbers below
```
exaexa, line 13: this probably shouldn't be here? (or is just a TODO? in that case pls name reasonably and mark it TODO. Or maybe show how to simulate crowding with smoment?)

stelmo: yep, just a place holder, I didn't want to rename all the files... But the TODO part is in the name: `10_ADD_ANOTHER_EXAMPLE.jl`

exaexa: pls leave it as 10_crowding and write that we should add how to simulate crowding, we'll solve this with the rest of docs later

stelmo: will do

PR #736, opened 2023-02-09
Q-opt support

exaexa:


src/solver.jl
```diff
@@ -37,7 +37,13 @@ function make_optimization_model(
...
     optimization_model = Model(optimizer)
     @variable(optimization_model, x[1:n])
-    @objective(optimization_model, sense, objective(model)' * x)
+    let obj = objective(model)
+        if obj isa AbstractVector
+            @objective(optimization_model, sense, obj' * x)
+        else
+            @objective(optimization_model, sense, x' * obj * [x; 1])
+        end
+    end
```
stelmo, line 46: This would cover 99% of use cases but not if someone has a weird objective like x^2 + x

exaexa: yeah actually that's covered there, the extra column is precisely for the `+x`.

exaexa: e.g. if you have 2 vars and want to do x[1]^2+x[1], you basically do
objective(m) = [1 0 1; 0 0 0]

stelmo: slightly harder to construct though?

exaexa: ah yes there's gonna be a tool for that

src/analysis/minimize_metabolic_adjustment.jl
```diff
@@ -34,75 +35,11 @@ value.(solution[:x])  # extract the flux from the optimizer
...
 minimize_metabolic_adjustment_analysis(
     model::AbstractMetabolicModel,
-    flux_ref::Union{Dict{String,Float64},Vector{Float64}},
+    reference_flux::Union{Vector{Float64}},
     optimizer;
-    modifications = [],
     kwargs...,
 ) = flux_balance_analysis(
-    model,
+    model |> Reconstruction.Pipes.minimize_adjustment(reference_flux),
```
stelmo, line 42: nice

src/solver.jl
```diff
@@ -37,7 +37,13 @@ function make_optimization_model(
 
     optimization_model = Model(optimizer)
     @variable(optimization_model, x[1:n])
-    @objective(optimization_model, sense, objective(model)' * x)
+    let obj = objective(model)
+        if obj isa AbstractVector
+            @objective(optimization_model, sense, obj' * x)
+        else
+            @objective(optimization_model, sense, x' * obj * [x; 1])
```
stelmo, line 44: todo later: explain this in the docs

src/types.jl
```diff
@@ -81,6 +80,7 @@ end # module Accessors
 end
 
 @inject Types.Internal begin
+    # TODO where is this declared?
```
stelmo, line 83: don't ask questions you don't want the answer to :)

src/types/wrappers/MinimizeAdjustment.jl
```diff
@@ -0,0 +1,30 @@
...
+adjusted_model = m |> MinimizeAdjustmentModel(fill(0.1, n_variables(m)))
+```
+"""
+struct MinimizeAdjustmentModel <: AbstractModelWrapper
+    inner::AbstractMetabolicModel
+    reference_assignment::Vector{Float64}
+end
+#TODO: sparse MinimizeAdjustmentModel with dict/sparseVec
+#TODO: MinimizeReactionAdjustmentModel?
+#TODO: MinimizeEnzymeAdjustmentModel?
```
stelmo, line 25: hmm it would be better if this were automatically handled, e.g. if you specify a variable in `reference_assignment` then it goes into the QP, otherwise it is ignored

stelmo (approved):
Small comment on generalizing the MOMA thing to automatically handle enzyme constrained models etc. Not necessary to implement now

src/types/wrappers/MinimizeAdjustment.jl
exaexa: Yeah I assume this could work by specifying the semantics directly. Will see when I'm at it.

PR #739, opened 2023-02-11
Add more reconstruction functions to ObjectModel

stelmo:
Adds non-inplace reconstruction functions and pipes for ObjectModel. Looking at the code, it screams "use a macro", but let's do this after 2.0 is out


exaexa (approved):
Pretty much OK.

This is creating a nice view on the reconstruction module where we group all the functions that modify the model structures without wrapping.

PR #740, opened 2023-02-11
Remove switching argument for reaction constructors

stelmo:


exaexa (approved):
OK.

I feel this will get a slight rename when we do the docs (`ReactionForwardOnly` or so).

PR #742, opened 2023-02-11
Add some more semantics to EnzymeConstrainedModel

stelmo:
Semantics are awesome ❤️

### DIFF
--- a/src/solver.jl
+++ b/src/solver.jl
@@ -120,6 +120,17 @@ solved_objective_value(flux_balance_analysis(model, ...))
 solved_objective_value(opt_model)::Maybe{Float64} =
     is_solved(opt_model) ? objective_value(opt_model) : nothing
 
+"""
+$(TYPEDSIGNATURES)
+
+From the optimized model, returns a vector of values for the selected
+`semantics`. If the model did not solve, returns `nothing`.
+
+# Example
+```
+values(Val(:reaction), model, flux_balance_analysis(model, ...)) # in order of reactions(model)
+```
+"""
 function values(
     semantics::Val{Semantics},
     model::AbstractMetabolicModel,
@@ -131,9 +142,31 @@ function values(
     is_solved(opt_model) ? sem_varmtx(model)' * value.(opt_model[:x]) : nothing
 end
 
+"""
+$(TYPEDSIGNATURES)
+
+Convenience variant of [`values`](@ref).
+
+# Example
+```
+values(:reaction, model, flux_balance_analysis(model, ...)) # in order of reactions(model)
+```
+"""
 values(semantics::Symbol, model::AbstractMetabolicModel, opt_model) =
     values(Val(semantics), model, opt_model)
 
+"""
+$(TYPEDSIGNATURES)
+
+From the optimized model, returns a dictionary mapping semantic IDs to their
+solved values for the selected `semantics`. If the model did not solve, returns
+`nothing`.
+
+# Example
+```
+values_dict(Val(:reaction), model, flux_balance_analysis(model, ...))
+```
+"""
 function values_dict(
     semantics::Val{Semantics},
     model::AbstractMetabolicModel,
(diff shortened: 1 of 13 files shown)


src/solver.jl
```diff
@@ -130,9 +148,7 @@ Returns a vector of fluxes of the model, if solved.
 flux_vector(flux_balance_analysis(model, ...))
 ```
 """
-flux_vector(model::AbstractMetabolicModel, opt_model)::Maybe{Vector{Float64}} =
-    is_solved(opt_model) ? reaction_variables_matrix(model)' * value.(opt_model[:x]) :
-    nothing
+flux_vector(model::AbstractMetabolicModel, opt_model)::Maybe{Vector{Float64}} = get_solution(:reaction, model, opt_model)
```
exaexa, line 151: maybe we could just rename or completely eradicate this now? `flux_` sounds unsystematic to `reaction_`. Flux can be anything

src/solver.jl
```diff
@@ -120,6 +120,24 @@ solved_objective_value(flux_balance_analysis(model, ...))
...
+
+function get_solution_dict(semantics::Val{Semantics}, model::AbstractMetabolicModel, opt_model) where {Semantics}
+    sem = Accessors.Internal.get_semantics(semantics)
+    isnothing(sem) && throw(DomainError(semantics, "Unknown semantics"))
+    (ids, _, _, sem_varmtx) = sem
+    is_solved(opt_model) ? Dict(ids .=> sem_varmtx(model)' * value.(opt_model[:x])) : nothing
+end
+
+get_solution_dict(semantics::Symbol, model::AbstractMetabolicModel, opt_model) = get_solution_dict(Val(semantics), model, opt_model)
+
```
exaexa, line 140: This is great, thanks.

As a tuning opportunity, the `get_` prefixes implies an imperative extraction operation, which should be possibly avoided, we want the whole thing to feel as stateless as possible. `solution` is probably too short, maybe we go for something like `solution_variable_values` and `solution_reaction_values` etc? Other ideas welcome tho.

src/types/wrappers/EnzymeConstrainedModel.jl
```diff
@@ -181,13 +181,23 @@ the original fluxes in the wrapped model
...
         variables(model),
-        semantics(model),
+        reactions(model),
         reaction_variables_matrix(model),
     ) # TODO currently inefficient
 
 """
 $(TYPEDSIGNATURES)
 
+Get a mapping of enzyme variables to variables - for enzyme constrained models,
```
exaexa, line 191:
```suggestion
Get a mapping of enzyme variables to variables -- for enzyme constrained models,
```
hyphen vs dash

src/types/wrappers/EnzymeConstrainedModel.jl
```diff
@@ -181,13 +181,23 @@ the original fluxes in the wrapped model
...
+this is just a direct mapping.
+"""
+Accessors.enzyme_variables(model::EnzymeConstrainedModel) = Dict(
+    gid => Dict(gid => 1.0) for gid in genes(model)
```
exaexa, line 195: oh lovely

src/solver.jl
```diff
@@ -120,6 +120,24 @@ solved_objective_value(flux_balance_analysis(model, ...))
...
     is_solved(opt_model) ? objective_value(opt_model) : nothing
 
+function get_solution(semantics::Val{Semantics}, model::AbstractMetabolicModel, opt_model) where {Semantics}
+    sem = Accessors.Internal.get_semantics(semantics)
+    isnothing(sem) && throw(DomainError(semantics, "Unknown semantics"))
+    (_, _, _, sem_varmtx) = sem
+    is_solved(opt_model) ? sem_varmtx(model)' * value.(opt_model[:x]) : nothing
+end
+
+get_solution(semantics::Symbol, model::AbstractMetabolicModel, opt_model) = get_solution(Val(semantics), model, opt_model)
```
exaexa, line 130: for this to be pipeable you need to remove the last argument and return a new closure
```suggestion
get_solution(semantics::Symbol, model::AbstractMetabolicModel) = opt_model -> get_solution(Val(semantics), model, opt_model)
```

src/solver.jl
```diff
@@ -120,6 +120,24 @@ solved_objective_value(flux_balance_analysis(model, ...))
...
+
+function get_solution_dict(semantics::Val{Semantics}, model::AbstractMetabolicModel, opt_model) where {Semantics}
+    sem = Accessors.Internal.get_semantics(semantics)
+    isnothing(sem) && throw(DomainError(semantics, "Unknown semantics"))
+    (ids, _, _, sem_varmtx) = sem
+    is_solved(opt_model) ? Dict(ids .=> sem_varmtx(model)' * value.(opt_model[:x])) : nothing
```
exaexa, line 136:
```suggestion
    is_solved(opt_model) ? Dict(ids(model) .=> sem_varmtx(model)' * value.(opt_model[:x])) : nothing
```
`ids` is a getter function

exaexa: (this should probably fix the tests)

exaexa (commented):
I'm perfectly OK with this once the tests work (see the comments for hints)

src/solver.jl
stelmo: will think of something (currently favoring `value`)

src/solver.jl
stelmo: will eradicate this in the next PR

exaexa (approved):
VERY GOOD.

I stil think we will need to find a good name for the `values_*`.... perhaps even `solution_values_...` ? but that can be left for later.

PR #749, opened 2023-02-12
Add another convenience constructor for enzyme constrained models

stelmo:
Adds a simpler fallback for `make_enzyme_constrained_model` and generalizes the pipe. Also added the test.


exaexa (approved):
Hey btw isn't this one already isomorphic to the "simplified" enzyme-constrained version? Code removal opportunity shouldn't be missed. :D

PR #750, opened 2023-02-13
Add model annotation details

stelmo:
Since ObjectModel is the go-to model type, it should contain as much info as possible. The latest community standards paper actually lists some of the requirements that models should possess. Much of this information can be included already into the annotations fields for Reactions, Metabolites and Genes. However, the model lacks these fields. Here I just co-opt these for the model. The only drawback is that `Annotations` is just a dictionary, and to be miriam compliant it should be a vector of tuples like `Vector{Tuple{String, String, String}}`, I think. 

See paper: https://www.embopress.org/doi/full/10.15252/msb.20199235


exaexa (approved):
oh lovely

exaexa:
PS it would be useful to probably make accessors for that, model_annotations and model_notes is OK. Not sure if that encodes to JSON but SBML should be very much OK

stelmo:
cool, I will add accessors too!

exaexa (changes requested):
- conversion function is discarding the information, can we fix it?
- I'll add this to SBMLModel (which actually supports this) and merge into this PR
- Does JSONModel support this?

stelmo:
So unfortunately JSON model is not really standardized, I am not sure where they would store this kind of information. They do have an `id` field at the top level of the object, but that is about it....

stelmo:
I just added what I could for JSONModel

stelmo:
It is failing, but this is probably due to an SBML.jl problem

PR #747, opened 2023-02-12
pFBA pipe

stelmo:
Adds a basic pipe-able pFBA variant. Should play nicely with more interesting model types, e.g. enzyme constrained etc.
```julia
gm_changed_bound =
    model |>
    with_changed_bound("BIOMASS_Ecoli_core_w_GAM", lower_bound = growth_lb) |>
    with_enzyme_constraints(
        gene_product_mass_group_bound = Dict(
            "uncategorized" => total_gene_product_mass,
        ),
    )

pfba_sol =
  gm_changed_bound |>
  with_parsimonious_solution(:enzyme) |>
  flux_balance_analysis(Clarabel.Optimizer)
```
@exaexa what do you think?

PR #751, opened 2023-02-13
Add pipes to all the analysis functions

stelmo:
I want this to be a reality:
```julia
model |> with_changed_bounds(...) |> with_added_reaction(...) |> with_enzyme_constraints(...) |> flux_balance_analysis(...) |> to_dict(:enzymes)
```


stelmo:
Related to #747

exaexa:
This is good. Do we add the Result overloads here?

stelmo:
Yes! I was thinking that each analysis function returns both the input model and the solved jump model (probably wrapped up into a type)

PR #753, opened 2023-02-15
remove and clean up GeneAssociations alias

exaexa:
- it was too similar to GeneAssociationsDNF
- Vector{Isozyme} actually explains the contents very well, and no complexity is hidden


stelmo (approved):
I agree!

PR #756, opened 2023-02-16
Export MIN_SENSE and MAX_SENSE

stelmo:


exaexa (commented):
This is creating a name conflict and a duality that we might want to avoid. If we really want to have the MAX_SENSE,MIN_SENSE exported, better name them differently and have our own types (so that the problem from jump2.0 doesn't ever happen again)

stelmo:
Hmm maybe the best solution is to just document it better...

stelmo:
closing this PR, will document at later stage

PR #757, opened 2023-02-16
Return enzyme concentrations in useful units

stelmo:


src/types/wrappers/EnzymeConstrainedModel.jl
```diff
@@ -189,15 +189,17 @@ Accessors.reaction_variables(model::EnzymeConstrainedModel) =
 $(TYPEDSIGNATURES)
 
 Get a mapping of enzyme variables to variables -- for enzyme constrained models,
-this is just a direct mapping.
+this is proportional to the molar mass of each gene product. Effectively the
```
exaexa, line 192: "to the molar mass of each gene product" is misleading ("molar mass" means "mass per mol" which it is not, and people will ignore the "proportional to" because it's not specified how.)

perhaps say directly

" In EnzymeConstrainedModel(@ref), this represents total mass influx of the enymes" or something.

exaexa (commented):
2 questions:
- SimplifiedEnzymeConstrainedModel does not need the same treatment?
- do we want to enforce the "enzyme's mass" semantics for the whole enzyme semantics?

stelmo:
So simplified enzyme does not have protein variables, so this is not needed there.
Not sure what you mean for the second question?

exaexa (approved):

PR #762, opened 2023-02-19
Fix and extend reconstruction functions and pipes for objectmodel

stelmo:
adds non-inplace `change_objective` as well as pipe
fixes `change_bound` to copy all fields
adds `change_gene_product_bound` in-place, non-inplace, as well as pipes
Also change modifications to have the prefix `modify_` instead of `change`. This was to prevent name space issues.

Reconstruction functions have `add_`, `remove_` and `change_` prefixes now.

### DIFF
--- a/src/misc/check_keys.jl
+++ b/src/misc/check_keys.jl
@@ -4,7 +4,8 @@ Throw an ArgumentError if model.field has any keys in xs::Union{Reaction, Metabo
 function throw_argerror_if_key_found(model, field, xs)
     _ids = collect(keys(getfield(model, field)))
     ids = [x.id for x in xs if x.id in _ids]
-    isempty(ids) || throw(ArgumentError("Duplicated $field IDs already present in model: $ids"))
+    isempty(ids) ||
+        throw(ArgumentError("Duplicated $field IDs already present in model: $ids"))
     nothing
 end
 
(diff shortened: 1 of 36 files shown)


stelmo:
No idea why this is failing...

stelmo:
This turned into a big housekeeping session

stelmo:
@exaexa the formatter causes this PR to break because it tries to split lines incorrectly. 
Before formatting:
```julia
@io_log @warn "Automatically converting $(typeof(model)) to JSONModel for saving, information may be lost."
```
After formatting (error):
```julia
@io_log
@warn "Automatically converting $(typeof(model)) to JSONModel for saving, information may be lost."
```

stelmo:
@exaexa I think this should be ready now

src/misc/check_keys.jl
```diff
@@ -0,0 +1,29 @@
+"""
+Throw an ArgumentError if model.field has any keys in xs::Union{Reaction, Metabolite, Gene}.
+"""
+function throw_argerror_if_key_found(model, field, xs)
```
exaexa, line 4: name: `arg_keys_exists`

use:
`arg_keys_exists(model.field, ks)`

src/misc/check_keys.jl
```diff
@@ -0,0 +1,29 @@
...
+    _ids = collect(keys(getfield(model, field)))
+    ids = [x.id for x in xs if x.id in _ids]
```
exaexa, line 6: collecting the keyset is suuuuper slow, use `haskey` !

src/misc/check_keys.jl
```diff
@@ -0,0 +1,29 @@
...
+    isempty(ids) ||
+        throw(ArgumentError("Duplicated $field IDs already present in model: $ids"))
```
exaexa, line 8: error: DomainError(ks, "keys already in model")`

exaexa (changes requested):
Mostly okay, small tech changes needed

src/misc/check_keys.jl
stelmo: need `field` to generate the nice error message.

PR #764, opened 2023-02-20
Fix cFBA implementation

stelmo:


stelmo:
this PR allows this kind of functionality:
```julia
cm = CommunityModel(
    members = [cm1, cm2],
    abundances = [a1, a2],
    environmental_links = [
        EnvironmentalLink(rid, mid, lb, ub) for
        (rid, mid, lb, ub) in zip(ex_rxns, ex_mids, ex_lbs, ex_ubs)
    ],
)

cm |> 
    with_changed_environmental_bound(...) |> 
    with_changed_abundances(...)
    with_equal_growth_objective() |>
    flux_balance_analysis(...; modifications =[modify_abundances(...)]) |>
    values_dict(:environmental_reaction)

and some other convenience functions
```

src/types/accessors/AbstractMetabolicModel.jl
```diff
@@ -155,6 +155,16 @@ negative (2nd law of thermodynamics). This semantic grouping represents ΔGᵣ.
...
+@make_variable_semantics(
+    :environmental_reaction,
+    "Environmental reaction",
+    """
+Community models are composed of member models as well as environmental exchange
+reactions. This semantic grouping represents the environmental exchange
+reactions.
+"""
+)
+
```
exaexa, line 167: could we actually call this exchanges right away? :D

src/types/models/CommunityModel.jl
```diff
@@ -0,0 +1,228 @@
...
+    (:gene_notes, Dict()),
+    (:reaction_name, nothing),
+    (:metabolite_name, nothing),
+    (:gene_name, nothing),
+)
+    @eval begin
+        Accessors.$func(cm::CommunityModel, id::String) =
+            access_community_member(cm, id, $func; default = $def)
+    end
+end
```
exaexa, line 228: these should unmap the reaction ID to the community members and recurse to the contained models.

src/reconstruction/CommunityModel.jl
```diff
@@ -0,0 +1,116 @@
...
+"""
+$(TYPEDSIGNATURES)
+
+Return an [`EqualGrowthCommunityModel`](@ref) wrapper around `model`, optionally
+specifying the `community_objective_id`.
+"""
+make_EqualGrowthCommunityModel(
+    model::CommunityModel;
+    community_objective_id = "equal_growth_rates_biomass_function",
+) = EqualGrowthCommunityModel(model, community_objective_id)
```
exaexa, line 116: this is unnecessary, people can just omit the make_ prefix and use a smart constructor instead.

src/types/misc/CommunityModel.jl
```diff
@@ -0,0 +1,116 @@
...
+    accessor(cm.members[idx].model, string(last(id_split)))
+end
+
+
+"""
+$(TYPEDSIGNATURES)
+
+A helper function to add the id of the community member as a prefix to some string.
+"""
+add_community_prefix(m::CommunityMember, str::String; delim = "#") = m.id * delim * str
```
exaexa, line 116:
```suggestion
add_community_prefix(m::CommunityMember, str::String; delim = "/") = "$(m.id)$delim$str"
```

exaexa: the "#" or "/" should be a constant btw

src/types/misc/CommunityModel.jl
```diff
@@ -0,0 +1,116 @@
...
+    cm::CommunityModel,
+    id::String,
+    accessor::Function;
+    default = nothing,
+)
+    id_split = split(id, "#")
+    idx = findfirst(startswith(first(id_split)), m.id for m in cm.members)
+    isnothing(idx) && return default # can only access inside community member
+    accessor(cm.members[idx].model, string(last(id_split)))
+end
```
exaexa, line 108: - the "default" sounds dangerous, I'd just go for error :D
- `findfirst` when you actually should have a Dict of the members
- `startswith` is dangerous, should be equality
- possibly you don't need to split the whole thing, just the part before the separator (can be done without an allocation using a string view)

src/types/models/CommunityModel.jl
```diff
@@ -0,0 +1,228 @@
...
+
+# Assumptions
+1. Exchange reactions, *all* of which are idenitified in
+   `exchange_reaction_ids`, have the form: `A[external] ⟷ ∅` where `A` is a
+   metabolite. No other exchanges are allowed. This is not checked, but assumed.
+2. There is only one biomass reaction in the model.
+"""
+Base.@kwdef mutable struct CommunityMember
+    "Name of model appended to intracellular reactions and metabolites."
+    id::String
```
exaexa, line 18: externalize ID

src/types/models/CommunityModel.jl
```diff
@@ -0,0 +1,228 @@
...
+    [`CommunityMember`](@ref) appended as a prefix with the delimiter `#`.
+3. All genes have the `id` of the respective underlying
+    [`CommunityMember`](@ref) appended as a prefix with the delimiter `#`.
+4. `environmental_links` is a superset of all exchange reactions contained in
+    each underlying member in the community. Only these reactions get joined to
+    each underlying model.
+"""
+Base.@kwdef mutable struct CommunityModel <: AbstractMetabolicModel
+    "Models making up the community."
+    members::Vector{CommunityMember}
```
exaexa, line 88: this should be a Dict

exaexa: (or there should be a lookup structure for members)

src/utils/CommunityModel.jl
```diff
@@ -0,0 +1,47 @@
...
+
+    (sem_ids, _, sem_vard, _) = Accessors.Internal.semantics(semantics)
+
+    ids = sem_ids(community_member.model)
+
+    Dict(
+        id => sum(v * val_d[k] for (k, v) in sem_vard(community_member.model)[id]) for
+        id in ids
+    )
+end
```
exaexa, line 26: - the sem_vard dict must be only constructed once (it's pricey)

also I think this is not a completely good way to approach this thing. Could we instead have a function like split_community_names and filter_member or so which allows the user to construct this themselves easily? As in, the function this way looks pretty good but it's very single-purpose-ish, and spends a lot of code on redoing stuff that could (should!) be done using the basic accessors.

Ideally, it could be
```
... |> values_dict(:reactions) |> split_community_members_dict
```
gets you
```
Dict(organism 1 => (reaction 1 => ..., reaction 2 => ...), organism 2 => ...)
```

or better
```
... |> filter_community_model_dict
```

src/types/misc/CommunityModel.jl
```diff
@@ -0,0 +1,116 @@
...
+    env_rxns::Vector{String},
+)
+    mat = spzeros(length(env_mets), n_variables(m.model))
+    for (midx, mid) in enumerate(env_mets)
+        mid in metabolites(m.model) || continue # does not have this exchange
+        ridx = first(indexin([env_rxns[midx]], variables(m.model))) #  find index of exchange reaction in member
+        mat[midx, ridx] = 1.0
+    end
+    return mat
+end
```
exaexa, line 54: updating sparse matrices in the middle is ultra slow, use the helper function from semantics

src/types/wrappers/EqualGrowthCommunityModel.jl
```diff
@@ -0,0 +1,101 @@
...
+function Accessors.stoichiometry(cm::EqualGrowthCommunityModel)
+
+    S = stoichiometry(cm.inner)
+    obj_col = spzeros(size(S, 1))
+
+    obj_links = spzeros(length(cm.inner.members), size(S, 2))
+    for i = 1:length(cm.inner.members)
+        m = cm.inner.members[i]
+        j = first(
+            indexin([add_community_prefix(m, m.biomass_reaction_id)], variables(cm.inner)),
```
exaexa, line 47: this is quadratic, make a proper lookup table (or use the helper function from semantics to construct the whole matrix)

also, do not update sparse matrices from the middle, it's even more slow

exaexa (commented):
The structure is okay, there are a few minor things in the comments but generally this should go in.

src/types/models/CommunityModel.jl
exaexa: (re lookup, look at how JSONModel handles it)

src/reconstruction/CommunityModel.jl
stelmo: yeah, I just wanted to fit in with `make_enzyme_constrained...`. removing

src/types/accessors/AbstractMetabolicModel.jl
stelmo: yeah, but that is even more verbose: `environmental_exchange_reaction`, I wanted to save myself the typing

src/types/misc/CommunityModel.jl
stelmo: we use `#` for the enzyme constrained, why change it to `\`?

stelmo: I could make the delimiter a constant, but would have to change that for all the delimiters. I will add this to our todo list for 2.0 and do it later

src/types/wrappers/EqualGrowthCommunityModel.jl
stelmo: what do you mean here?

stelmo: this will have very few updates, just one for each biomass function

src/types/misc/CommunityModel.jl
stelmo: what do you mean?

src/types/wrappers/EqualGrowthCommunityModel.jl
stelmo: fixed this

src/types/misc/CommunityModel.jl
stelmo: I don't think the semantics would work on a community member here, rather just building the sparse matrix better

src/types/misc/CommunityModel.jl
stelmo: mostly did this, just not sure how to do the view

src/types/models/CommunityModel.jl
stelmo: still not sure what you mean here

src/types/misc/CommunityModel.jl
stelmo: Need the "default" here because accessors for variables like environmental exchanges need to default to something sensible.

src/utils/CommunityModel.jl
stelmo: mostly done with this, but need `bind`

src/types/misc/CommunityModel.jl
```diff
@@ -0,0 +1,136 @@
...
+    )
+end
+
+"""
+$(TYPEDSIGNATURES)
+
+A helper function that creates the entire exchange/environmental variable
+linking matrix for a community model.
+"""
+function env_ex_matrix(cm)
```
exaexa, line 66: type (also, can we have a sensible name like environment_exchange_stoichiometry or so)

src/types/misc/CommunityModel.jl
```diff
@@ -0,0 +1,136 @@
...
+    )
+end
+
+"""
+$(TYPEDSIGNATURES)
+
+Variant of [`env_ex_matrix`](@ref) that takes an explicit abundance matrix (used
+in solver modifications.)
+"""
+function env_ex_matrix(cm, abundances)
```
exaexa, line 84: type

src/types/misc/CommunityModel.jl
```diff
@@ -0,0 +1,136 @@
...
+    upper_bound = m.upper_bound,
+)
+
+"""
+$(TYPEDSIGNATURES)
+
+A helper function that creates an exchange/environmental variable linking matrix
+for community member `m`.
+"""
+function env_ex_member_matrix(
```
exaexa, line 42: maybe extend the name a bit (see lower)

src/types/misc/CommunityModel.jl
```diff
@@ -0,0 +1,136 @@
...
+    delim_id::String,
+    accessor::Function;
+    delim = "#",
+    default = nothing,
+)
+    modelid_nameid = string.(split(delim_id, delim))
+    length(modelid_nameid) == 1 && return default # accessor default
+
+    modelid = first(modelid_nameid)
+    nameid = last(modelid_nameid)
```
exaexa, line 113: this assumes there is not going to be anything with more than 2 parts. Explicitly check that the length is >=2 and take [1] and the rest.

src/types/wrappers/EqualGrowthCommunityModel.jl
```diff
@@ -0,0 +1,97 @@
...
+
+# Notes
+1. No biomass metabolite exists (and none are created).
+
+# Fields
+$(TYPEDFIELDS)
+"""
+Base.@kwdef mutable struct EqualGrowthCommunityModel <: AbstractModelWrapper
+    inner::CommunityModel
+    community_objective_id::String = "community_biomass"
```
exaexa, line 17: is it objective id or reaction id?

test/reconstruction/ObjectModel.jl
```diff
@@ -185,4 +185,11 @@
...
     @test !isnothing(new_model2.reactions["r2"].gene_associations)
     @test isnothing(new_model.reactions["r2"].gene_associations)
+
+    # test added biomass metabolite
+    # new_model = model |> with_added_biomass_metabolite("BIOMASS_Ecoli_core_w_GAM")
+    # @test "biomass" in metabolites(modded_ecoli)
+    # @test !("biomass" in metabolites(ecoli))
+    # @test haskey(modded_ecoli.reactions["BIOMASS_Ecoli_core_w_GAM"].metabolites, "biomass")
+    # @test !haskey(ecoli.reactions["BIOMASS_Ecoli_core_w_GAM"].metabolites, "biomass")
 end
```
exaexa, line 195: this can be removed?

test/types/CommunityModel.jl
```diff
@@ -0,0 +1,416 @@
...
+            0.0 0.0 0.0 1.0 -1.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0
+            0.0 0.0 0.0 0.0 0.0 0.0 0.0 -1.0 -1.0 0.0 0.0 0.0 0.0
+            0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 1.0 -1.0 0.0 0.0 0.0
+            0.0 0.0 0.0 0.0 0.0 -1.0 0.0 0.0 0.0 1.0 0.0 0.0 0.0
+            0.0 0.0 0.0 0.0 0.0 1.0 -1.0 0.0 0.0 0.0 0.0 0.0 0.0
+            0.2 0.0 0.0 0.0 0.0 0.0 0.0 0.8 0.0 0.0 -1.0 0.0 0.0
+            0.0 0.0 0.0 0.0 0.2 0.0 0.0 0.0 0.0 0.0 0.0 -0.2 0.0
+            0.0 0.0 0.0 0.0 0.0 0.0 0.8 0.0 0.0 0.0 0.0 0.0 -0.8
+        ],
+    )
```
exaexa, line 129: There are no dicts in the way so this doesn't depend on dict order, right?

exaexa (commented):
minor technical stuff, otherwise good to go

test/types/CommunityModel.jl
stelmo: nope, I purposefully used either vectors or `OrderedDict`s to make this work

src/types/wrappers/EqualGrowthCommunityModel.jl
stelmo: this is a special kind of variable, not really a reaction in this context

src/types/misc/CommunityModel.jl
stelmo: Fixed

exaexa (approved):
btw FFTM when the type annotations are fixed

src/types/misc/CommunityModel.jl
stelmo: this is actually much more problematic, we need a better way to map through from a delimited ID to the base ID, and also do this recursively.

stelmo: Made this an issue in #765

stelmo:
Fixed the types issues and renamed the cryptic methods. Did not fix the delimiter issue, will need tlc in #765

PR #766, opened 2023-02-24
Add basic show for ModelWithResult

stelmo:
```julia
julia> res = flux_balance_analysis(model, Tulip.Optimizer)
ModelWithResult{Model}(COBREXA.Types.ObjectModel, ...)
```

### DIFF
--- a/src/io/show/ModelWithResult.jl
+++ b/src/io/show/ModelWithResult.jl
@@ -1,0 +1,5 @@
+function Base.show(io::Base.IO, ::MIME"text/plain", res::ModelWithResult)
+    println(io, "A ModelWithResult composed of:")
+    println(io, "model: $(typeof(res.model))")
+    println(io, "result: $(typeof(res.result))")
+end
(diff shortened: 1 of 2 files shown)


src/io/show/ModelWithResult.jl
```diff
@@ -0,0 +1,5 @@
+function Base.show(io::Base.IO, ::MIME"text/plain", res::ModelWithResult)
+    println(io, "A ModelWithResult composed of:")
+    println(io, "model: $(typeof(res.model))")
+    println(io, "result: $(typeof(res.result))")
+end
```
exaexa, line 5: show should be preferably as parseable as possible, multiline is discouraged.

Let's go for this:
```suggestion
function Base.show(io::Base.IO, ::MIME"text/plain", res::ModelWithResult{T}) where T
    println(io, "ModelWithResult{$T}($(res.model), ...)")
end
```

exaexa (changes requested):
suggestion not tested tho

exaexa (approved):

PR #767, opened 2023-02-27
split wrappers off to a separate module

exaexa:


src/reconstruction/enzyme_constrained.jl
```diff
@@ -1,3 +1,4 @@
+# TODO move this to wrappers
```
exaexa, line 1: @stelmo I thought that this "generic" reconstruction code really should be in wrappers (together with the models). It technically _is_ reconstruction but in pure reconstruction I'd say we'd have just the traditional construction/reconstruction stuff.

stelmo: agreed!

stelmo (approved):
This is great! 🚀

PR #773, opened 2023-03-03
implement wrappy pFBA using the normal L2 wrappers

exaexa:


exaexa:
@stelmo I've got 1 remaining test fail on this, which is actually on ECModel (the result is like 89% instd of 91% of some enzyme). Could you pls check what might have gone wrong? Otherwise this is IMO gtg

stelmo:
Will look now, could be the solver misbehaving though

stelmo (approved):

PR #776, opened 2023-05-02
make the semantics cut

exaexa:
I have this specially separated from the rest of the (RELATIVELY HARMLESS) semantics-are-everywhere change because it's gonna be the most breaking part, and the reviews should be kinda clear.

In this state it's not looking good (tests will fail until I fix them) but most of the errors I see seem fixable.

As the main points:
- `stoichiometry` is now just a nice name for `metabolite_variables_matrix`
- `balance` is gone, instead of that the metabolites have the balance bounds
- stoichiometry is no more matrixy "by default", the dictionary-ish `metabolite_variables` is the master form now as with other semantics..
  - pros: resizing variables or reactions doesn't kill stoichiometry and you don't need to care about that much
  - cons: reconstructing the stoichiometry from dicts isn't that nicely efficient, we'll need a bit of extra annotations to be able to pipe the matrixy form through wrappers easily
  
I'll fix the rest, at this points any comments welcome.

### DIFF
--- a/docs/src/concepts/4_wrappers.md
+++ b/docs/src/concepts/4_wrappers.md
@@ -108,7 +108,7 @@ modifying the reaction list, stoichiometry, and bounds:
 COBREXA.unwrap_model(x::LeakyModel) = x.mdl
 COBREXA.reaction_count(x::LeakyModel) = reaction_count(x.mdl) + 1
 COBREXA.reaction_ids(x::LeakyModel) = [reaction_ids(x.mdl); "The Leak"]
-COBREXA.stoichiometry(x::LeakyModel) = [stoichiometry(x.mdl) [m in x.leaking_metabolites ? -1.0 : 0.0 for m = metabolites(x.mdl)]]
+COBREXA.stoichiometry(x::LeakyModel) = [stoichiometry(x.mdl) [m in x.leaking_metabolites ? -1.0 : 0.0 for m = metabolite_ids(x.mdl)]]
 function COBREXA.variable_bounds(x::LeakyModel)
     (l, u) = variable_bounds(x.mdl)
     return ([l; x.leak_rate], [u; x.leak_rate])
--- a/src/analysis/reconstruction/gapfill_minimum_reactions.jl
+++ b/src/analysis/reconstruction/gapfill_minimum_reactions.jl
@@ -54,7 +54,7 @@ function gapfill_minimum_reactions(
     precache!(model)
 
     # constraints from universal reactions that can fill gaps
-    univs = _universal_stoichiometry(universal_reactions, metabolites(model))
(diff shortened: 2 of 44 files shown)


docs/src/concepts/4_wrappers.md
```diff
@@ -108,7 +108,7 @@ modifying the reaction list, stoichiometry, and bounds:
 COBREXA.unwrap_model(x::LeakyModel) = x.mdl
 COBREXA.reaction_count(x::LeakyModel) = reaction_count(x.mdl) + 1
 COBREXA.reaction_ids(x::LeakyModel) = [reaction_ids(x.mdl); "The Leak"]
-COBREXA.stoichiometry(x::LeakyModel) = [stoichiometry(x.mdl) [m in x.leaking_metabolites ? -1.0 : 0.0 for m = metabolites(x.mdl)]]
+COBREXA.stoichiometry(x::LeakyModel) = [stoichiometry(x.mdl) [m in x.leaking_metabolites ? -1.0 : 0.0 for m = metabolite_ids(x.mdl)]]
```
stelmo, line 111: probably want a better name than `leaking_metabolites`

src/analysis/reconstruction/gapfill_minimum_reactions.jl
```diff
@@ -54,7 +54,7 @@ function gapfill_minimum_reactions(
     precache!(model)
 
     # constraints from universal reactions that can fill gaps
-    univs = _universal_stoichiometry(universal_reactions, metabolites(model))
+    univs = _universal_stoichiometry(universal_reactions, metabolite_ids(model))
```
stelmo, line 57: Love it

src/analysis/reconstruction/gapfill_minimum_reactions.jl
```diff
@@ -89,7 +89,7 @@ function gapfill_minimum_reactions(
     @constraint(
         opt_model,
         extended_stoichiometry * [x; ux] .==
-        [balance(model); zeros(length(univs.new_mids))]
+        [metabolite_bounds(model); zeros(length(univs.new_mids))]
```
stelmo, line 92: YES

stelmo (approved):
YES

exaexa:
@stelmo I guess this is ready for merging into the bigger PR now (at least the tests pass). It might be necessary to redo the logic of the community models a bit. Matrix passthrough still doesn't really work but that's patchable later (it's mostly a minor performance issue)

stelmo:
> @stelmo I guess this is ready for merging into the bigger PR now (at least the tests pass). It might be necessary to redo the logic of the community models a bit. Matrix passthrough still doesn't really work but that's patchable later (it's mostly a minor performance issue)

🚀 let's figure out the problems as they come up

exaexa:
ok :boom:

PR #780, opened 2023-05-05
Throw error if too many arguments supplied to enzyme models

stelmo:


src/reconstruction/enzyme_constrained.jl
```diff
@@ -42,6 +42,15 @@ function make_enzyme_constrained_model(
...
+        !isnothing,
+        [
+            gene_product_mass_groups,
+            gene_product_mass_group_bounds,
+            total_gene_product_mass_bound,
+        ],
+    ) && throw(ArgumentError("Too many arguments specified!"))
+
     if !isnothing(total_gene_product_mass_bound)
         gene_product_mass_groups = Dict("uncategorized" => genes(model))
```
exaexa, line 55: the condition actually belongs here, here you want to check that gene_product_mass_groups is indeed nothing before the assignment. OTOH if you tripped over this I guess we should just completely erase the third "too default" argument and replace it with some function that makes the enzymes "all uncategorized"

stelmo: I actually want to remove all the default behavior. The user must specify at least something and it needs to be non-contradictory

exaexa (commented):
Any idea why this fails the tests? looks like constraints missing or something.

src/reconstruction/simplified_enzyme_constrained.jl
```diff
@@ -44,15 +44,22 @@ function make_simplified_enzyme_constrained_model(
...
     if !isnothing(total_reaction_mass_bound)
-        reaction_mass_groups = Dict("uncategorized" => variables(model)) # TODO should be reactions
+        (isnothing(reaction_mass_groups) && isnothing(reaction_mass_group_bounds)) ||
+            throw(
+                ArgumentError(
+                    "argument values would be overwritten by total_reaction_mass_bound!",
+                ),
+            )
+
+        reaction_mass_groups = Dict("uncategorized" => genes(model))
```
exaexa, line 56: bug spotted: why is it genes here when it is reaction_mass_group?

exaexa (changes requested):

src/reconstruction/simplified_enzyme_constrained.jl
exaexa: (in turn, this generates a coupling matrix without entries (because nothing mathes the reaction ids))

exaexa:
```suggestion
        reaction_mass_groups = Dict("uncategorized" => variables(model))
```

exaexa (approved):
oh finally all ok

PR #783, opened 2023-05-11
fix sparsevec

exaexa:


stelmo (approved):

exaexa:
@stelmo there are 2 other sparsevec instances in community models, can you have a look if they actually do what they are supposed to do? The problem is that `sparsevec([],[],...)` has defaulted to a sparse vector of `Any`s, which don't have a default zero, thus allocation of anything breaks.

PR #786, opened 2023-05-18
Extend the SBO identifiers and add isa_xxx functions

stelmo:


src/utils/looks_like.jl
```diff
@@ -140,10 +140,14 @@ forwarded to [`looks_like_extracellular_metabolite`](@ref).
...
-@_is_reaction_fn "pseudo" Identifiers.PSEUDOREACTIONS
-@_is_reaction_fn "metabolic" Identifiers.METABOLIC_REACTIONS
-@_is_reaction_fn "spontaneous" Identifiers.SPONTANEOUS_REACTIONS
+@_is_sbo_reaction_fn "exchange" Identifiers.EXCHANGE_REACTIONS
+@_is_sbo_reaction_fn "transport" Identifiers.TRANSPORT_REACTIONS
+@_is_sbo_reaction_fn "biomass" Identifiers.BIOMASS_REACTIONS
+@_is_sbo_reaction_fn "atp_maintenance" Identifiers.ATP_MAINTENANCE_REACTIONS
+@_is_sbo_reaction_fn "pseudo" Identifiers.PSEUDO_REACTIONS
+@_is_sbo_reaction_fn "metabolic" Identifiers.METABOLIC_REACTIONS
+@_is_sbo_reaction_fn "spontaneous" Identifiers.SPONTANEOUS_REACTIONS
```
exaexa, line 149: I'd vote for using symbols here right away, as in:
```suggestion
@_is_sbo_reaction_fn :spontaneous Identifiers.SPONTANEOUS_REACTIONS
```

not a big deal but prevents weird errors like leaving a space in the string (which doesn't translate very well to function names :D )

exaexa (approved):
lovely, consider the 2 comments (both minor nitpicking)

src/utils/looks_like.jl
stelmo: Cool, I will do this, but in FBCMTs.

stelmo:
Note, all of this sbo stuff will also go to FBCMTs, so after this merge, expect a PR cutting it out :)

PR #787, opened 2023-05-19
Make duplicate reaction test stricter

stelmo:
Takes into account reaction direction now


exaexa (approved):
lgtm

(this is the thing that might be eventually moved to fbcmt, right?)

stelmo:
Yep, I will :ax it shortly

PR #790, opened 2023-06-09
Neaten the summarize functions

stelmo:


stelmo:
Also added exchange semantics to make this prettier

stelmo:
@exaexa can you update your CI julia distributions please? they are running a Julia from the stone ages

Project.toml
```diff
@@ -24,9 +25,8 @@ StableRNGs = "860ef19b-820b-49d6-a774-d7a799459cd3"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [compat]
-Clarabel = "0.3"
```
exaexa, line 27: how do we test QPs if clarabel is removed?

src/types/accessors/bits/semantics.jl
```diff
@@ -239,3 +239,23 @@ macro all_variables_are_reactions(mt)
...
+            vid for vid in $Accessors.variables(model) if
+            length(keys($Accessors.reaction_stoichiometry(model, vid))) == 1
+        ]
+        $Accessors.n_exchanges(model::$m) = length($Accessors.exchanges(model))
+        $Accessors.exchange_variables(model::$m) = Dict(
+            vid => Dict(vid => 1.0) for vid in $Accessors.variables(model) if
+            length(keys($Accessors.reaction_stoichiometry(model, vid))) == 1
+        )
+    end
+end
```
exaexa, line 261: lovely

exaexa:
One extra wish, Can we split off separate PRs with the exchanges and PrettyTables so that the commits aren't interwimbgled?

stelmo:
> Collaborator

Unfortunately not, I did not follow a good coding pattern doing this

Project.toml
stelmo: I don't think it is necessary in the compat section, it is in the extras section. will see if anything breaks

exaexa: extras section does not list the version, which is the main problem. This doesn't technically break anything but the breakage will appear once clarabel grows.

(Also the package without bounds will get rejected by General. :D )

exaexa:
obsoleted by CTs

PR #800, opened 2023-11-10
Start getting FBA to work

stelmo:
Fixed some compat issues, and deleted modifications. Get FBA to run and test it

### DIFF
--- a/docs/src/examples/02-flux-balance-analysis.jl
+++ b/docs/src/examples/02-flux-balance-analysis.jl
@@ -7,7 +7,6 @@ import JSONFBCModels as J
 
 model = A.load(J.JSONFBCModel, "e_coli_core.json")
 
-
 ctmodel = X.fbc_model_constraints(model)
 
 
--- a/docs/src/examples/02-flux-balance-analysis.jl
+++ b/docs/src/examples/02-flux-balance-analysis.jl
@@ -4,9 +4,15 @@
 import COBREXA as X
 import AbstractFBCModels as A
 import JSONFBCModels as J
+import ConstraintTrees as C
+using Gurobi
 
 model = A.load(J.JSONFBCModel, "e_coli_core.json")
 
 ctmodel = X.fbc_model_constraints(model)
 
+vt = C.ValueTree(ctmodel, value.(opt_model[:x]))
 
+vt = X.flux_balance_analysis(model, Gurobi.Optimizer)
+
+vt = X.flux_balance_analysis(ctmodel, Gurobi.Optimizer; modifications=[X.silence])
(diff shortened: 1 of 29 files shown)


exaexa:
pls pls rebase

docs/src/examples/02-flux-balance-analysis.jl
```diff
@@ -4,9 +4,15 @@
 import COBREXA as X
 import AbstractFBCModels as A
 import JSONFBCModels as J
+import ConstraintTrees as C
+using Gurobi
```
exaexa, line 8: proprietary

src/COBREXA.jl
```diff
@@ -17,6 +17,7 @@ include("builders/core.jl")
 include("builders/genes.jl")
 include("builders/objectives.jl")
 
+include("analysis/modifications/optimizer.jl")
```
exaexa, line 20: we should rename this a bit, something like solver settings or so

src/analysis/flux_balance_analysis.jl
```diff
@@ -51,4 +61,6 @@ $(TYPEDSIGNATURES)
 Pipe-able variant of [`flux_balance_analysis`](@ref).
 """
 flux_balance_analysis(optimizer; modifications = []) =
-    model -> flux_balance_analysis(model, optimizer; modifications)
+    m -> flux_balance_analysis(m, optimizer; modifications)
+
+export flux_balance_analysis
```
exaexa, line 66: vscode settiiiings

src/analysis/modifications/optimizer.jl
```diff
@@ -0,0 +1,36 @@
+
+"""
+$(TYPEDSIGNATURES)
+
+Change the objective sense of optimization. Possible arguments are
+`JuMP.MAX_SENSE` and `JuMP.MIN_SENSE`.
+"""
+modify_sense(objective_sense) =
```
exaexa, line 8: we don't need a modification nomenclature here so maybe we can even have the same names as JuMP or so, to prevent chaos.

Too bad julia doesn't have partial application, that would be exactly it

exaexa (commented):
good

docs/src/examples/02-flux-balance-analysis.jl
stelmo: already gone :)

src/COBREXA.jl
stelmo: will do

src/analysis/flux_balance_analysis.jl
stelmo: Huh?

src/analysis/modifications/optimizer.jl
stelmo: aahhh

stelmo: okay

exaexa:
ok I guess let's shoot. Thanks @stelmo !

PR #802, opened 2023-11-11
Implement quadratic functionality

stelmo:


exaexa:
@stelmo if you could have a quick look now it would be great. I've set up proper intermediate (CT-based) functions for FBA/pFBA; same will follow with MOMA and FVA and likely later for others. These will also be used exclusively for the complex models.

I split off the MOMA from QP tutorial. To be more realistic I assume we generally want to run pFBA instead of plain FBA to obtain a sane initial "reference" solution, right?

docs/src/examples/03-parsimonious-flux-balance.jl
```diff
@@ -0,0 +1,99 @@
...
+
+ctmodel = fbc_model_constraints(model)
+ctmodel *= :l2objective^squared_sum_objective(ctmodel.fluxes)
+ctmodel.objective.bound = 0.3 # set growth rate # TODO currently breaks
+
+opt_model = optimization_model(
+    ctmodel;
+    objective = ctmodel.:l2objective.value,
+    optimizer = Clarabel.Optimizer,
+    sense = Minimal,
```
stelmo, line 49: I don't like this `Minimal`. Since we are using JuMP's `set_optimizer_attribute` etc., we should use JuMP's `Min` and `Max`

docs/src/examples/03-parsimonious-flux-balance.jl
```diff
@@ -0,0 +1,99 @@
...
+)
+
+J.optimize!(opt_model) # JuMP is called J in COBREXA
+
+is_solved(opt_model) # check if solved
+
+vt = C.constraint_values(ctmodel, J.value.(opt_model[:x])) # ConstraintTrees.jl is called C in COBREXA
+
+@test isapprox(vt.l2objective, ?; atol = QP_TEST_TOLERANCE) #src  # TODO will break until mutable bounds
```
stelmo, line 58: Can remove TODO

docs/src/examples/03-parsimonious-flux-balance.jl
```diff
@@ -0,0 +1,99 @@
...
```
stelmo, line 97: Can remove TODO

src/analysis/modifications.jl
```diff
@@ -4,8 +4,8 @@
 """
 $(TYPEDSIGNATURES)
 
-Change the objective sense of optimization. Possible arguments are
-`JuMP.MAX_SENSE` and `JuMP.MIN_SENSE`.
+Change the objective sense of optimization. Accepted arguments include
+[`Minimal`](@ref), [`Maximal`](@ref), and [`Feasible`](@ref).
```
stelmo, line 8: What does `Feasible` do?

src/solver.jl
```diff
@@ -47,3 +47,30 @@ is_solved(opt_model::J.Model) =
...
+const Maximal = J.MAX_SENSE
+
+"""
+    Maximal
+
+Objective sense for finding the any feasible value of the objective.
+
+Same as `JuMP.FEASIBILITY_SENSE`.
+"""
+const Feasible = J.FEASIBILITY_SENSE
```
stelmo, line 76: Did not know this exists

src/solver.jl
```diff
@@ -47,3 +47,30 @@ is_solved(opt_model::J.Model) =
...
+"""
+const Maximal = J.MAX_SENSE
+
+"""
+    Maximal
+
+Objective sense for finding the any feasible value of the objective.
+
+Same as `JuMP.FEASIBILITY_SENSE`.
+"""
```
stelmo, line 75: My concern here is that it is just more stuff to remember. Can we not just re-export the JuMP constants?

stelmo (commented):
Small things

stelmo:
> @stelmo if you could have a quick look now it would be great. I've set up proper intermediate (CT-based) functions for FBA/pFBA; same will follow with MOMA and FVA and likely later for others. These will also be used exclusively for the complex models.
> 
> I split off the MOMA from QP tutorial. To be more realistic I assume we generally want to run pFBA instead of plain FBA to obtain a sane initial "reference" solution, right?

Yep, it would make sense to use a pFBA solution for MOMA, but as long as its general i.e. the user can decide how sane they want to be, I don't mind

docs/src/examples/03-parsimonious-flux-balance.jl
exaexa: we're renaming JuMP's symbols to ours (people don't import them from JuMP!) so we shouldn't also force them to import parameters.

docs/src/examples/03-parsimonious-flux-balance.jl
exaexa: This is still commented out :D

src/analysis/modifications.jl
exaexa: well right, see docs :D

https://github.com/LCSB-BioCore/COBREXA.jl/pull/802/commits/53cc6b75fd7e8b01e9e97fec99746dcc4b3f2f40#diff-9180cb5ed3696043617bf7efd6fbca6dc96db4c3b8186d07b1851e5d2edbd95fR72

https://jump.dev/JuMP.jl/stable/api/JuMP/#JuMP.OptimizationSense

src/solver.jl
exaexa: one can use the jump constants as well. I just don't want to 1] cause identifier collision 2] have folks to import jump all over again everytime they want to do something

exaexa:
> Yep, it would make sense to use a pFBA solution for MOMA, but as long as its general i.e. the user can decide how sane they want to be, I don't mind

I converted this to an admonition opportunity.

PR #808, opened 2023-12-16
Add max min driving force capability

stelmo:
Tested against v1, and they generate the same solution.

### DIFF
--- a/src/builders/thermodynamic.jl
+++ b/src/builders/thermodynamic.jl
@@ -1,0 +1,144 @@
+"""
+The function uses the supplied `optimizer` and `reaction_standard_gibbs_free_energies`.
+Optionally, `flux_solution` can be used to set the directions of each reaction in `model`
+(all reactions are assumed to proceed forward and are active by default). The supplied
+`flux_solution` should be free of internal cycles i.e. thermodynamically consistent. This
+optional input is important if a reaction in `model` normally runs in reverse (negative
+flux). Note, reactions in `flux_solution` that are smaller than `small_flux_tol` are also
+ignored in the analysis function (for numerical stability).
+
+
+Reactions specified in `ignore_reaction_ids` are internally ignored when calculating the
+max-min driving force. This should include water and proton importers.
+
+Since biochemical thermodynamics are assumed, the `proton_ids` and `water_ids` need to be
+specified so that they can be ignored in the calculations. Effectively this assumes an
+aqueous environment at constant pH is used.
+
+`constant_concentrations` is used to fix the concentrations of certain metabolites (such as
+CO₂). `concentration_ratios` is used to specify additional constraints on metabolite pair
+concentrations (typically, this is done with various cofactors such as the ATP/ADP ratio.
+For example, you can fix the concentration of ATP to be always 5× higher than of ADP by
+specifying `Dict(("ATP","ADP") => 5.0)`
+
+`concentration_lb` and `concentration_ub` set the `Cₗ` and `Cᵤ` in the
+optimization problems.
+
+`T` and `R` can be specified in the corresponding units; defaults are K and kJ/K/mol.
+"""    
+function max_min_driving_force(
+    model::MetabolicModel,
+    reaction_standard_gibbs_free_energies::Dict{String,Float64},
+    optimizer;
+    flux_solution::Dict{String,Float64} = Dict{String,Float64}(),
+    proton_ids::Vector{String} = ["h_c", "h_e"],
+    water_ids::Vector{String} = ["h2o_c", "h2o_e"],
+    constant_concentrations::Dict{String,Float64} = Dict{String,Float64}(),
+    concentration_ratios::Dict{Tuple{String,String},Float64} = Dict{
+        Tuple{String,String},
+        Float64,
+    }(),
+    concentration_lb = 1e-9,
+    concentration_ub = 100e-3,
+    T = _constants.T,
+    R = _constants.R,
+    small_flux_tol = 1e-6,
+    modifications = [],
+    ignore_reaction_ids = [],
+)
+    opt_model = Model(optimizer)
+
+    @variables opt_model begin
+        mmdf
+        logcs[1:n_metabolites(model)]
+        dgrs[1:n_reactions(model)]
+    end
+
+    # set proton log concentration to zero so that it won't impact any calculations (biothermodynamics assumption)
+    proton_idxs = Int.(indexin(proton_ids, metabolites(model)))
+    for idx in proton_idxs
+        JuMP.fix(logcs[idx], 0.0)
+    end
+
+    # set water concentrations to zero (aqueous condition assumptions)
+    water_idxs = Int.(indexin(water_ids, metabolites(model)))
+    for idx in water_idxs
+        JuMP.fix(logcs[idx], 0.0)
+    end
+
+    # only consider reactions with supplied thermodynamic data AND have a flux bigger than
+    # small_flux_tol => finds a thermodynamic profile that explains flux_solution
+    active_rids = filter(
+        rid ->
+            haskey(reaction_standard_gibbs_free_energies, rid) &&
+                abs(get(flux_solution, rid, small_flux_tol / 2)) > small_flux_tol &&
+                !(rid in ignore_reaction_ids),
+        reactions(model),
+    )
+    active_ridxs = Int.(indexin(active_rids, reactions(model)))
+
+    # give dummy dG0 for reactions that don't have data
+    dg0s =
+        [get(reaction_standard_gibbs_free_energies, rid, 0.0) for rid in reactions(model)]
+
+    S = stoichiometry(model)
+
+    @constraint(opt_model, dgrs .== dg0s .+ (R * T) * S' * logcs)
+
+    # thermodynamics should correspond to the fluxes
+    flux_signs = [sign(get(flux_solution, rid, 1.0)) for rid in reactions(model)]
+
+    # only constrain reactions that have thermo data
+    @constraint(opt_model, dgrs[active_ridxs] .* flux_signs[active_ridxs] .<= 0)
+
+    # add the absolute bounds
+    missing_mets =
+        [mid for mid in keys(constant_concentrations) if !(mid in metabolites(model))]
+    !isempty(missing_mets) &&
+        throw(DomainError(missing_mets, "metabolite(s) not found in model."))
+    for (midx, mid) in enumerate(metabolites(model)) # idx in opt_model (missing ignore_metabolites)
+        midx in water_idxs && continue
+        midx in proton_idxs && continue
+        if haskey(constant_concentrations, mid)
+            JuMP.fix(logcs[midx], log(constant_concentrations[mid]))
+        else
+            # this metabolite needs bounds
+            @constraint(
+                opt_model,
+                log(concentration_lb) <= logcs[midx] <= log(concentration_ub)
+            )
+        end
+    end
+
+    # add the relative bounds
+    for ((mid1, mid2), val) in concentration_ratios
+        idxs = indexin([mid1, mid2], metabolites(model)) # TODO: this is not performant
+        any(isnothing.(idxs)) &&
+            throw(DomainError((mid1, mid2), "metabolite pair not found in model."))
+        @constraint(opt_model, logcs[idxs[1]] == log(val) + logcs[idxs[2]])
+    end
+
+    @constraint(opt_model, mmdf .<= -dgrs[active_ridxs] .* flux_signs[active_ridxs])
+
+    @objective(opt_model, Max, mmdf)
+
+    # apply the modifications, if any
+    for mod in modifications
+        mod(model, opt_model)
+    end
+
+    optimize!(opt_model)
+
+    is_solved(opt_model) || return nothing
+
+    return (
+        mmdf = value(opt_model[:mmdf]),
+        dg_reactions = Dict(
+            rid => value(opt_model[:dgrs][i]) for (i, rid) in enumerate(reactions(model))
+        ),
+        concentrations = Dict(
+            mid => exp(value(opt_model[:logcs][i])) for
+            (i, mid) in enumerate(metabolites(model))
+        ),
+    )
+end
--- a/src/builders/thermodynamic.jl
+++ b/src/builders/thermodynamic.jl
@@ -136,26 +136,17 @@ m = add_ratio_constraints(
 )
 ```
 """
-function add_log_ratio_constraints(
-    m::C.ConstraintTree,
-    ratios::Dict{String,Tuple{String,String,Float64}};
-    name::Symbol,
+log_ratio_constraints(
+    ratios::Dict{String,Tuple{String,String,Float64}},
     on::C.ConstraintTree,
+) = C.ConstraintTree(
+    Symbol(cid) => C.Constraint(
+        value = on[Symbol(var1)].value - on[Symbol(var2)].value,
+        bound = log(ratio),
+    ) for (cid, (var1, var2, ratio)) in ratios
 )
-    for (cid, (var1, var2, ratio)) in ratios
-        m *=
-            name^Symbol(
-                cid,
-            )^C.Constraint(
-                value = on[Symbol(var1)].value - on[Symbol(var2)].value,
-                bound = log(ratio),
-            )
-    end
 
-    m
-end
-
-export add_log_ratio_constraints
+export log_ratio_constraints
 
 
 """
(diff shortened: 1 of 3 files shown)


exaexa:
I guess all good, can we merge into the enzyme branch? That one is also good, I'll just need to go through that.

src/builders/thermodynamic.jl
```diff
@@ -0,0 +1,258 @@
...
+        concentration_ratios,
+    )
+    
+    optimized_constraints(
+        m;
+        objective = m.max_min_driving_force.value,
+        optimizer,
+        modifications,
+    )
+end
```
exaexa, line 258: pls vscode config

src/builders/thermodynamic.jl
```diff
@@ -0,0 +1,258 @@
...
+    ignore_reaction_ids = String[],
+)
+    # check if reactions that will be used in the model all have thermodynamic data, otherwise throw error.
+    # only these reactions will form part of the model.
+    rxns = filter(
+        x -> !(x in ignore_reaction_ids),
+        isempty(flux_solution) ? A.reactions(model) : collect(keys(flux_solution)),
+    )
+    all(in.(rxns, Ref(collect(keys(reaction_standard_gibbs_free_energies))))) || throw(
+        ArgumentError("""
```
exaexa, line 54: this is DomainError

src/builders/thermodynamic.jl
```diff
@@ -0,0 +1,258 @@
...
+    R = 8.31446261815324e-3, # kJ/K/mol
+    ignore_reaction_ids = String[],
+)
+    # check if reactions that will be used in the model all have thermodynamic data, otherwise throw error.
+    # only these reactions will form part of the model.
+    rxns = filter(
+        x -> !(x in ignore_reaction_ids),
+        isempty(flux_solution) ? A.reactions(model) : collect(keys(flux_solution)),
+    )
+    all(in.(rxns, Ref(collect(keys(reaction_standard_gibbs_free_energies))))) || throw(
```
exaexa, line 53: use set difference (this is N^2)

src/builders/thermodynamic.jl
```diff
@@ -0,0 +1,258 @@
...
+    concentration_lb = 1e-9,
+    concentration_ub = 100e-3,
+    T = 298.15, # Kelvin
+    R = 8.31446261815324e-3, # kJ/K/mol
+    ignore_reaction_ids = String[],
+)
+    # check if reactions that will be used in the model all have thermodynamic data, otherwise throw error.
+    # only these reactions will form part of the model.
+    rxns = filter(
+        x -> !(x in ignore_reaction_ids),
```
exaexa, line 50: use Set (this is N^2)

src/builders/thermodynamic.jl
exaexa: ...on a side thought, why not just consider the reactions that have thermo data instead of forcing people to make sure that every reaction that doesn't have the thermo data gets filled precisely into the ignored ids?

src/builders/thermodynamic.jl
```diff
@@ -0,0 +1,258 @@
...
+concentration of ATP to be always 5× higher than of ADP by specifying
+`Dict(("ATP","ADP") => 5.0)`
+
+`concentration_lb` and `concentration_ub` set default concentration bounds, in M
+by default.
+
+`T` and `R` can be specified in the corresponding units; defaults are K and
+kJ/K/mol.
+"""
+function build_max_min_driving_force_model(
```
exaexa, line 35: naming: this is fbc_max_min_driving_force_constraints()

(we might also rename the fbc_model_constraints  to fbc_steady_flux_constraints or so)

src/builders/thermodynamic.jl
```diff
@@ -0,0 +1,258 @@
...
+The function uses the supplied `optimizer` and
+`reaction_standard_gibbs_free_energies`. Optionally, `flux_solution` can be used
+to set the directions of each reaction in `model` (all reactions are assumed to
+proceed forward and are active by default). The supplied `flux_solution` should
+be free of internal cycles i.e. thermodynamically consistent. This optional
+input is important if a reaction in `model` normally runs in reverse (negative
+flux). Note, reactions in `flux_solution` that are smaller than `small_flux_tol`
+are also ignored in the analysis function (for numerical stability).
+
+
```
exaexa, line 13: double nl?

src/builders/thermodynamic.jl
```diff
@@ -0,0 +1,258 @@
...
+metabolites (such as CO₂). `concentration_ratios` is used to specify additional
+constraints on metabolite pair concentrations (typically, this is done with
+various cofactors such as the ATP/ADP ratio. For example, you can fix the
+concentration of ATP to be always 5× higher than of ADP by specifying
+`Dict(("ATP","ADP") => 5.0)`
+
+`concentration_lb` and `concentration_ub` set default concentration bounds, in M
+by default.
+
+`T` and `R` can be specified in the corresponding units; defaults are K and
```
exaexa, line 32: "the provided defaults are in K and ..."

src/builders/thermodynamic.jl
```diff
@@ -0,0 +1,258 @@
...
+CO₂). `concentration_ratios` is used to specify additional constraints on metabolite pair
+concentrations (typically, this is done with various cofactors such as the ATP/ADP ratio.
+For example, you can fix the concentration of ATP to be always 5× higher than of ADP by
+specifying `Dict(("ATP","ADP") => 5.0)`
+
+`concentration_lb` and `concentration_ub` set the `Cₗ` and `Cᵤ` in the
+optimization problems.
+
+`T` and `R` can be specified in the corresponding units; defaults are K and kJ/K/mol.
+"""
```
exaexa, line 218: docstrings too long (who's gonna maintain two copies :D ), better refer to the fact that the arguments are forwarded. Also helps to have the defaults only in a single place.

src/builders/thermodynamic.jl
```diff
@@ -0,0 +1,258 @@
...
+function max_min_driving_force_analysis(
+    model::A.AbstractFBCModel,
+    reaction_standard_gibbs_free_energies::Dict{String,Float64};
+    flux_solution = Dict{String,Float64}(),
+    concentration_ratios = Dict{String,Tuple{String,String,Float64}}(),
+    proton_ids = ["h_c", "h_e"],
+    water_ids = ["h2o_c", "h2o_e"],
+    concentration_lb = 1e-9,
+    concentration_ub = 100e-3,
```
exaexa, line 227: this is unnecessarily confusing and incomparable to the above `1e-9`. It's simply `1e-1`. With `100e-3` people need to go (`-3+2`) to actually see the magnitude.

src/builders/thermodynamic.jl
```diff
@@ -0,0 +1,258 @@
...
+                      """),
+    )
+
+    # import metabolite ids (if used), and reaction stoichiometries from an AbstractFBCModel
+    mets = Set(Symbol[])
+    for rid in rxns
+        for met in keys(A.reaction_stoichiometry(model, rid))
+            push!(mets, Symbol(met))
+        end
+    end
```
exaexa, line 67:
```suggestion
    mets = Set(m for m in keys(A.reaction_stoichiometry(model, rid)) for rid in rxns)
```

src/builders/thermodynamic.jl
```diff
@@ -0,0 +1,258 @@
...
+    stoi = A.stoichiometry(model)
+
+    # create thermodynamic variables
+    m = C.ConstraintTree(:max_min_driving_force^C.variable())
+    m +=
+        :log_metabolite_concentrations^C.variables(
+            keys = mets,
+            bounds = zip(log(concentration_lb), log(concentration_ub)),
+        )
+    m += :delta_G_reactions^C.variables(keys = Symbol.(rxns))
```
exaexa, line 78: why not do in 1 go?
```suggestion
    m = C.ConstraintTree(
        :max_min_driving_force => C.variable(),
        :log_metabolite_concentrations => ....
     )
```

src/builders/thermodynamic.jl
```diff
@@ -0,0 +1,258 @@
...
+    end
+
+    #= 
+    Set proton log concentration to zero so that it won't impact any
+    calculations (biothermodynamics assumption). Also set water concentrations
+    to zero (aqueous condition assumptions). How true is "aqueous conditions"? 
+    Debatable... 
+    =#
+    log_met_conc_zero = [Symbol.(proton_ids); Symbol.(water_ids)]
+    for met in keys(m.log_metabolite_concentrations)
```
exaexa, line 115: this unnecessarily checks the log_met_conc_zero many times (and that check is slow), better swap the cycle order (because check for key membership is easy). Also might be more easy to just throw an error if people don't have the IDs correct.

src/builders/thermodynamic.jl
```diff
@@ -0,0 +1,258 @@
...
+
+`T` and `R` can be specified in the corresponding units; defaults are K and
+kJ/K/mol.
+"""
+function build_max_min_driving_force_model(
+    model::A.AbstractFBCModel,
+    reaction_standard_gibbs_free_energies::Dict{String,Float64};
+    flux_solution = Dict{String,Float64}(),
+    proton_ids = ["h_c", "h_e"],
+    water_ids = ["h2o_c", "h2o_e"],
```
exaexa, line 40:
```suggestion
    proton_ids,
    water_ids,
```

avoid model-specific defaults

src/builders/thermodynamic.jl
```diff
@@ -0,0 +1,258 @@
...
+    =#
+    model_rxns = A.reactions(model)
+    model_mets = A.metabolites(model)
+    for rxn in rxns
+        met_idxs, stoich_coeffs = findnz(stoi[:, findfirst(==(rxn), model_rxns)])
+        met_ids = model_mets[met_idxs]
+
+        dG0 = reaction_standard_gibbs_free_energies[rxn]
+
+        m *=
```
exaexa, line 92: this mutates the thing million times (one per each reaction). Why run a loop when the dict can be constructed directly roughly as:

```
m *= :delta_g_reaction_equations ^ C.ConstraintTree(rid => .... for rid=rxns)
```
also please take care about rxns vs. rids naming

src/builders/thermodynamic.jl
```diff
@@ -0,0 +1,258 @@
...
+    m += :delta_G_reactions^C.variables(keys = Symbol.(rxns))
+
+    #= 
+    Build gibbs free energy relations
+    ΔG_rxn == ΔG0 + R * T * sum ( log_concentration_variable * stoichiometry_value )
+    =#
+    model_rxns = A.reactions(model)
+    model_mets = A.metabolites(model)
+    for rxn in rxns
+        met_idxs, stoich_coeffs = findnz(stoi[:, findfirst(==(rxn), model_rxns)])
```
exaexa, line 87: `findfirst` is slow. Why cycle through `rxns` and try to reconstruct the order for `stoi`chiometry when you can cycle through the stoichiometry right away and simply drop reactions that are not in rxns?

src/builders/thermodynamic.jl
```diff
@@ -0,0 +1,258 @@
...
+        end
+    end
+
+    #= 
+    Add thermodynamic feasibility constraint (ΔG < 0 for a feasible reaction in flux direction).
+    Add objective constraint to solve max min problem.
+    =#
+    for rxn in rxns
+        m *=
+            :thermodynamic_feasibility^Symbol(
```
exaexa, line 127: naming: the constraints together ensure the thermodynamic feasibility, but their values will not really mean "thermodynamic feasibility".. More like leftover deltaG or so... `:reaction_delta_G_margins` ?

src/builders/thermodynamic.jl
```diff
@@ -0,0 +1,258 @@
...
+                rxn,
+            )^C.Constraint(
+                value = m.delta_G_reactions[Symbol(rxn)].value *
+                        sign(get(flux_solution, rxn, 1.0)),
+                bound = (-Inf, 0.0),
+            )
+
+        m *=
+            :max_min_constraints^Symbol(
```
exaexa, line 136: same here, think for the future when this is solved, people actually look at it and want to know what the value means. `:min_driving_force_margins` (if I got it right)?

src/builders/thermodynamic.jl
```diff
@@ -0,0 +1,258 @@
...
+        if met in log_met_conc_zero
+            m.log_metabolite_concentrations[met].bound = 0.0
+        end
+    end
+
+    #= 
+    Add thermodynamic feasibility constraint (ΔG < 0 for a feasible reaction in flux direction).
+    Add objective constraint to solve max min problem.
+    =#
+    for rxn in rxns
```
exaexa, line 125: again, do not use `*=` in a cycle. Simply construct the ConstraintTree in one go and attach it in place.

src/builders/thermodynamic.jl
```diff
@@ -0,0 +1,258 @@
+"""
+$(TYPEDSIGNATURES)
+
+The function uses the supplied `optimizer` and
```
exaexa, line 4: ...but there's no optimizer here?

src/builders/thermodynamic.jl
```diff
@@ -0,0 +1,258 @@
...
+`concentration_lb` and `concentration_ub` set default concentration bounds, in M
+by default.
+
+`T` and `R` can be specified in the corresponding units; defaults are K and
+kJ/K/mol.
+"""
+function build_max_min_driving_force_model(
+    model::A.AbstractFBCModel,
+    reaction_standard_gibbs_free_energies::Dict{String,Float64};
+    flux_solution = Dict{String,Float64}(),
```
exaexa, line 38: rename to `reference_flux` or `flux_direction_target` or so. The "solution" has no description of how it relates to the result in itself.

src/builders/thermodynamic.jl
```diff
@@ -0,0 +1,258 @@
...
+                bound = (-Inf, 0.0),
+            )
+    end
+
+    m
+end
+
+export build_max_min_driving_force_model
+
+function add_metabolite_ratio_constraints!(
```
exaexa, line 151: this does not allocate variables, so there's absolutely no reason for the inplace functionality.

Also you can generalize it to ratio_constraints; it may constraint anything (even fluxes)

exaexa (changes requested):
ok let's make this perfect.

stelmo:
Yep, let me just fix up the docs, then merge in there

src/builders/thermodynamic.jl
stelmo: To prevent silent footguns. This way is more explicit. We had it the other way around before, but I don't like the silence

src/builders/thermodynamic.jl
stelmo: Hmm I actually like all the functions that build CTs to start with build. It makes it more obvious what they do. Hence, I would actually change `fbc_model_constraints` to `build_fbc_model_constraints`

src/builders/thermodynamic.jl
stelmo: ah, the docs are out of date, upding them now

src/builders/thermodynamic.jl
stelmo: this is great

src/builders/thermodynamic.jl
stelmo: better your way

src/builders/thermodynamic.jl
stelmo: good

src/builders/thermodynamic.jl
stelmo: This is so much better!

src/builders/thermodynamic.jl
stelmo: aahh did not know :)

stelmo: but it is more clunky like that though

src/builders/thermodynamic.jl
stelmo: either way will require some sort of check if the `rxn` is in the thermo model. Typically (I think) `rxns` will be much smaller than `model_rxns`, so better to cycle through the smaller one?

stelmo:
boom 🚀 💥

stelmo:
So the only thing failing is my cowboy test in enzymes

exaexa (commented):
I guess we're good now, will apply more :axe: :nerd_face: when it gets to the enzyme pr

exaexa (approved):

exaexa:
ok good all errors are in the other MR already

PR #809, opened 2023-12-17
Homogenize names of functions

stelmo:
Added suffix `_analysis` to all the one-shot functions, so that it is more obvious that they are the "easy" mode functions. Also people expect `flux_balance_analysis` not `flux_balance`.

Added prefix `build_` to all functions that build new constraint trees. Again, I think this consistent naming pattern will help people when they do `COBREXA.<tab>` to figure out what to use and what not to use.

Moved modifications to misc.


stelmo:
will make a new PR, too much has changed since I made this one

PR #819, opened 2024-01-29
Showcase: enzyme communities

stelmo:
Showcase enzyme communities


exaexa:
side comment: we'll have to host the data in some better way, this is not sustainable.

If the stuff can be downloaded, let's download strictly. If not we have to make a separate hosting for that, or use dummy values.

exaexa:
I assume this is moved to https://github.com/COBREXA/COBREXA.jl/pull/14

PR #820, opened 2024-01-29
Small fix to mmdf

stelmo:


docs/src/examples/06-mmdf.jl
```diff
@@ -16,8 +16,6 @@
 
 # # Thermodynamic models
 
-using COBREXA
-
```
exaexa, line 20: this needs to be here, people will fail without copypasting this.

src/frontend/mmdf.jl
```diff
@@ -56,8 +56,8 @@ Following arguments are set optionally:
 """
 function max_min_driving_force_analysis(
     model::A.AbstractFBCModel,
-    reaction_standard_gibbs_free_energies::Dict{String,Float64};
-    reference_flux = Dict{String,Float64}(),
+    reaction_standard_gibbs_free_energies::Dict{String,Float64},
+    reference_flux::Dict{String,Float64};
```
exaexa, line 60: o nice

src/builders/mmdf.jl
```diff
@@ -0,0 +1,74 @@
...
+        )
+    stoi = A.stoichiometry(model)
+
+    constraints =
+        :log_concentrations^C.variables(keys = mets, bounds = concentration_bound.(mets)) +
+        :reactant_log_concentrations^C.variables(keys = rxns)
+
+    cs = C.ConstraintTree()
+
+    midx_lookup = Dict(indexin(mets, Symbol.(A.metabolites(model))) .=> 1:length(mets))
```
exaexa, line 56: this sounds horribly inefficient; does one actually need to suffer from the midx lookups here?

Why not keep the whole thing sparse and just avoid adding to the matrix if the reaction and/or metabolite isn't in the happy subset?

at least comment is required.

exaexa (changes requested):
please separate the actual fix from code moving.

docs/src/examples/06-mmdf.jl
stelmo: it appears later on

src/builders/mmdf.jl
stelmo: the problem is the new model will have different indices from the old one. So just keeping track of the indices is required to make sure the mapping is correct at the end. Otherwise you just default to having the entire old model in the new (ideally smaller) model

